PR #18, opened 2021-08-28
docs: update instructions for hacs default

alandtse:
No longer needed to be added as a custom repository


alandtse:
So it looks like we may have to rename the component to be picked up as default. I'm debating this since it's probably a breaking change for users.

alandtse:
Not necessary to go into default. If people want to use it, they're welcome to self add.

Issue #22, opened 2021-09-08
use mdi:car-connected rather than mdi-update for the update-switch


arallsopp:
I find the Tesla's use of the word 'update' confusing. It seems to expose:

- *update available* - a sensor indicating that there is an update for the car's software.
- *update-switch* - an entity controlling whether HA continues to poll the car.

I think it would be better if we used the "mdi:car-connected" icon for the update-switch, and considered renaming it to 'enable polling' or 'enable scans' (as 'scan' is the word used in the configuration).

I've tweaked my dashboard to show these values so that I'm no longer confused, but I wonder if others might benefit from making the change in the actual integration.

![image](https://user-images.githubusercontent.com/2462104/132490448-30761cec-b91d-4c21-8007-ca4ec41103d1.png)

alandtse:
This is a good point. Feel free to submit a PR if you want. We can just fix the display name without making it a breaking change.

arallsopp:
Good work, @alandtse. Thanks.

PR #24, opened 2021-09-09
Arallsopp update switch

arallsopp:
Relates to Issue #22 

## Straightforward changes:

- the icon for 'update switch' from mdi:update to mdi:car-connected.
- the text 'seconds between scans' to 'seconds between polling'.

## More complicated:
I think I've also changed the name and ID of the update switch so that it appears as 'polling switch' in the UI and logs. I'm wondering if that's actually a breaking change, as potentially people have automations or scripts that might target the update switch entity.

### DIFF
--- a/custom_components/tesla_custom/const.py
+++ b/custom_components/tesla_custom/const.py
@@ -25,7 +25,7 @@ ICONS = {
     "parking brake sensor": "mdi:car-brake-parking",
     "charger sensor": "mdi:ev-station",
     "charger switch": "mdi:battery-charging",
-    "update switch": "mdi:update",
+    "update switch": "mdi:car-connected",
     "maxrange switch": "mdi:gauge-full",
     "temperature sensor": "mdi:thermometer",
     "location tracker": "mdi:crosshairs-gps",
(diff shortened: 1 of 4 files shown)


arallsopp:
@alandtse, if the breaking change is a pain, feel free to omit [commit 6f1e598](https://github.com/arallsopp/tesla/commit/6f1e59833823c666e3305af9872e5ba419f46fea) from this pull request.

alandtse:
Thanks! I want to get us moved to our final login scheme but I'll plan to get this in when we do that. It will be a breaking change because of the switch rename but it'll make sense since we'll be making some major changes.

arallsopp:
Excellent. Thanks for the update.

This integration saved my bacon last night, btw, as we'd left our car charging at a free pod point whilst grocery shopping, and an automation clocked we'd forgotten it at 11pm. That would have been a very bad start to this morning's commute :)

alandtse:
Merged in by #30

arallsopp:
Thanks, @alandtse. Hope it's helped.

PR #26, opened 2021-09-10
Switch to refresh_token auth

omelhus:
This removes the proxy and decorator-stuff and uses the refresh token instead.

It's pretty much the PR I sent to HA with a few changes to take the existing changes into account.

Works on my machine.


omelhus:
I didn’t think to remove (or even run) the tests 🤦🏼‍♂️

alandtse (changes requested):
Thanks. Couple of things if you're able to, otherwise I can this weekend.

custom_components/tesla_custom/__init__.py
```diff
@@ -27,30 +40,49 @@
...
+                ): vol.All(cv.positive_int, vol.Clamp(min=MIN_SCAN_INTERVAL)),
+            }
+        )
+    },
+    extra=vol.ALLOW_EXTRA,
+)
+
 
 @callback
-def _async_save_tokens(hass, config_entry, access_token, refresh_token, expiration):
```
alandtse, line 38: Let's keep the expiration in the config_flow. Otherwise we will automatically refresh the token in the first 5 minutes after start up. Given how paranoid Tesla is, we want to minimize auth attempts.

custom_components/tesla_custom/__init__.py
```diff
@@ -132,15 +165,41 @@ async def async_setup_entry(hass, config_entry):
...
+    except IncompleteCredentials as ex:
+        await async_client.aclose()
+        raise ConfigEntryAuthFailed from ex
+    except httpx.ConnectTimeout as ex:
+        await async_client.aclose()
+        raise ConfigEntryNotReady from ex
     except TeslaException as ex:
+        await async_client.aclose()
         if ex.code == HTTP_UNAUTHORIZED:
-            _async_start_reauth(hass, config_entry)
```
alandtse, line 140: We need to keep the reauth code. Otherwise uses will have to remove and readd the integration every 45 days.

custom_components/tesla_custom/__init__.py
```diff
@@ -132,15 +165,41 @@ async def async_setup_entry(hass, config_entry):
...
+            "TOO_MANY_REQUESTS",
+            "SERVICE_MAINTENANCE",
+            "UPSTREAM_TIMEOUT",
+        ]:
+            raise ConfigEntryNotReady(
+                f"Temporarily unable to communicate with Tesla API: {ex.message}"
+            ) from ex
         _LOGGER.error("Unable to communicate with Tesla API: %s", ex.message)
         return False
-    _async_save_tokens(hass, config_entry, access_token, refresh_token, expiration)
```
alandtse, line 143: Not sure why we'd remove this, but perhaps it's being saved somewhere else?

custom_components/tesla_custom/__init__.py
```diff
@@ -162,27 +221,16 @@ async def async_setup_entry(hass, config_entry):
...
-        await asyncio.gather(
-            *[
-                hass.config_entries.async_forward_entry_unload(config_entry, platform)
-                for platform in PLATFORMS
-            ]
-        )
+    unload_ok = await hass.config_entries.async_unload_platforms(
+        config_entry, PLATFORMS
     )
-    await hass.data[DOMAIN].get(config_entry.entry_id)[
```
alandtse, line 183: The controller should get the disconnect signal if we're going down so it stops polling Tesla.

custom_components/tesla_custom/__init__.py
```diff
@@ -193,17 +241,6 @@ async def async_unload_entry(hass, config_entry) -> bool:
     return False
 
 
-def _async_start_reauth(hass: HomeAssistant, entry: ConfigEntry):
```
alandtse, line 196: See earlier statement about reauth.

custom_components/tesla_custom/__init__.py
```diff
@@ -240,9 +277,8 @@ async def _async_update_data(self):
             result = self.controller.get_tokens()
             refresh_token = result["refresh_token"]
             access_token = result["access_token"]
-            expiration = result["expiration"]
```
alandtse, line 243: Pass in the expiration per prior discussion.

custom_components/tesla_custom/__init__.py
```diff
@@ -240,9 +277,8 @@ async def _async_update_data(self):
...
             _async_save_tokens(
-                self.hass, self.config_entry, access_token, refresh_token, expiration
```
alandtse, line 245: Same.

custom_components/tesla_custom/__init__.py
```diff
@@ -251,7 +287,5 @@ async def _async_update_data(self):
             # handled by the data update coordinator.
             async with async_timeout.timeout(30):
                 return await self.controller.update()
-        except IncompleteCredentials:
```
alandtse, line 254: This probably should still be there in case the refresh token is wrong.

custom_components/tesla_custom/climate.py
```diff
@@ -80,15 +80,13 @@ def target_temperature(self):
         """Return the temperature we try to reach."""
         return self.tesla_device.get_goal_temp()
 
-    @TeslaDevice.Decorators.check_for_reauth
```
alandtse, line 83: We may have to double check in teslajsonpy but I think this should still be good and will help catch a reauth need.

custom_components/tesla_custom/config_flow.py
```diff
@@ -206,35 +141,40 @@ async def async_step_init(self, user_input=None):
...
+        result = await controller.connect(test_login=True)
+        config[CONF_TOKEN] = result["refresh_token"]
+        config[CONF_ACCESS_TOKEN] = result["access_token"]
+        config[CONF_EXPIRATION] = result[CONF_EXPIRATION]
+        config[CONF_USERNAME] = data[CONF_USERNAME]
+
+    except IncompleteCredentials as ex:
+        _LOGGER.error("Authentication error: %s %s", ex.message, ex)
+        raise InvalidAuth() from ex
     except TeslaException as ex:
```
alandtse, line 170: Need to keep IncompleteCredentials

custom_components/tesla_custom/config_flow.py
```diff
@@ -206,35 +141,40 @@ async def async_step_init(self, user_input=None):
...
-        if ex.code == HTTP_UNAUTHORIZED or isinstance(ex, IncompleteCredentials):
-            _LOGGER.error("Invalid credentials: %s", ex.message)
+        if ex.code == HTTP_UNAUTHORIZED:
+            _LOGGER.error("Invalid credentials: %s", ex)
```
alandtse, line 172: ex.message

custom_components/tesla_custom/lock.py
```diff
@@ -24,13 +24,11 @@ async def async_setup_entry(hass, config_entry, async_add_entities):
 class TeslaLock(TeslaDevice, LockEntity):
     """Representation of a Tesla door lock."""
 
-    @TeslaDevice.Decorators.check_for_reauth
```
alandtse, line 27: Same on decorators for reauth.

custom_components/tesla_custom/strings.json
```diff
@@ -12,8 +12,12 @@
     },
     "step": {
       "user": {
-        "description": "The next step will forward to the proxy for Tesla authentication. Tesla protects the login with a web application firewall (WAF) which may require a refresh to get or receive data. Please be cautious with multiple refreshes in a short period of time.",
-        "title": "Tesla - Configuration"
+        "description": "Use 'Auth App for Tesla' on iOS or 'Tesla Tokens' on Android\r\n to create a refresh token and enter it below.",
```
alandtse, line 15: I believe this is Markdown, do links work?

custom_components/tesla_custom/tesla_device.py
```diff
@@ -1,61 +1,25 @@
...
 
 from .const import DOMAIN, ICONS
 
 _LOGGER = logging.getLogger(__name__)
 
-
 class TeslaDevice(CoordinatorEntity):
     """Representation of a Tesla device."""
 
-    class Decorators(CoordinatorEntity):
```
alandtse, line 21: Restore per earlier discussion.

custom_components/tesla_custom/tesla_device.py
```diff
@@ -1,61 +1,25 @@
...
-
     def __init__(self, tesla_device, coordinator):
         """Initialise the Tesla device."""
         super().__init__(coordinator)
         self.tesla_device = tesla_device
-        self._name: str = self.tesla_device.name
-        self._unique_id: str = slugify(self.tesla_device.uniq_name)
-        self._attributes: str = self.tesla_device.attrs.copy()
-        self.config_entry_id: Optional[str] = None
+        self._name = self.tesla_device.name
```
alandtse, line 20: Let's keep the prior typing hints.

tests/test_tesla_device.py
```diff
@@ -2,7 +2,6 @@
 from unittest.mock import MagicMock, Mock
 
 import pytest
-from teslajsonpy.exceptions import IncompleteCredentials
```
alandtse, line 5: Restore.

tests/test_tesla_device.py
```diff
@@ -35,27 +34,6 @@ def tesla_device_mock(tesla_api_mock):
     return TeslaDevice(tesla_api_mock, coordinator)
 
 
-@pytest.fixture
```
alandtse, line 38: Restore.

custom_components/tesla_custom/lock.py
omelhus: You think we still need them?

custom_components/tesla_custom/strings.json
omelhus: Did not think to try markdown. I tried html and it borks.

omelhus:
Alright, so I won’t be able to do much until next week. Have at it this weekend and I’ll pick up if anything on Monday.

alandtse:
Thanks, I took your changes in manually in #29 as I wanted to double check where I had branched from core earlier. Thank you again for your help in getting this started.

omelhus:
Totally understand.

A known issue is that current version of core requires 0.18.3 of the teslajsonpi, and I’m not sure if it will upgrade to 0.20.0 if the Tesla component in core is active. This will result in a “invalid auth” error and left me confused for a bit. I fixed this by manually removing the package from site-packages.

This is of course no longer a problem as soon as the core component is gone.

alandtse:
If you add the custom component in the integration menu, it will overwrite it automatically.

PR #43, opened 2021-09-19
Specify platform for authentication apps

thatneat:
just a quick README update so that folks can more quickly get where they need to be.


alandtse:
Good call. Do you mind also doing it for info.md? That's shown in HACS.

alandtse:
Broken CI is unrelated to the docs change. I'll plan to merge this as soon as info.md is also updated.

PR #47, opened 2021-09-27
Add repository.json

dougfultz:
Add repository configuration per https://github.com/home-assistant/developers.home-assistant/blob/930ec445f61da4099c7de1f787028f41cdf7e8cb/docs/add-ons/repository.md

Fixes the error encountered when attempting to add repository.

### DIFF
--- a/repository.json
+++ b/repository.json
@@ -1,0 +1,5 @@
+{
+    "name": "Tesla Custom Integration",
+    "url": "https://github.com/alandtse/tesla",
+    "maintainer": "Alan D. Tse <alandtse@gmail.com>"
+}


alandtse:
Thanks. However, this is not an add-on and cannot be installed that way. We must be added via HACS as a custom integration.

dougfultz:
Worth a shot. 😄  It might be beneficial to add instructions to the readme on adding the custom integration repository.  My Googling seemed to indicate that adding integration repositories isn't in the UI anymore. :(

alandtse:
The readme already has those instructions. It's step 1.

dougfultz:
Step one only indicates to add it as a custom repository:

`Use HACS after adding this https://github.com/alandtse/tesla as a custom repository.`

For give my ignorance.  But I can't seem to locate *how* to add it.  I found https://community.home-assistant.io/t/hacs-add-custom-repository-in-settings/209574/14 but the workaround doesn't seem to work.

PR #56, opened 2021-10-09
Added Support for Energy Sites

giachello:
I added support for energy sites. This pull request goes together with 0.21.2 on teslajsonpy. I've tested it on my solar panels for a week or so and it works. To integrate the power sensor into Home Assistant's Energy panel, you need to integrate its value by adding an integration sensor like this:

```
sensor:
  - platform: integration
    source: sensor.my_tesla_solar_panel
    name: Solar Panel Energy Generated
    unit_prefix: k
    round: 2
````


giachello:
Ok, closed this pull request, will create a new one once teslajsonpy is done.

alandtse:
As a note, the teslajsonpy are breaking changes so we're iterating up to 1.0.0.

Issue #65, opened 2021-10-15
Integration documentation missing


trombonekenny:
Since HA landed your PR to remove the Tesla integration https://github.com/home-assistant/core/pull/55988, it also removed the documentation page linked in this project's README.

https://www.home-assistant.io/integrations/tesla/ no longer describes the integration. Is there any alternative source of those docs somewhere?

brendan41357:
I would also like this. I found general TeslaMate docs at https://docs.teslamate.org/docs/installation/docker which has a Home Assistant section, but it does not seem to give a lot of info.

For instance, I have no idea what half of the entities that were generated when I installed the integration are for, if there is a way to adjust polling intervals, or why 1/3 of the entities (such as battery sensor and temperature sensors) are "unknown"

arallsopp:
In the absence of anyone jumping on your comment above, I've opened a pull request (#94) to move the base information into the readme file. This would make it viewable inside HACS, and can always be used to point users towards the wiki if the documentation eventually expands.

Thanks as always for your work on this integration.

pvmil:
Would it be possible to add service call documentation on defrosting the car, or setting it to a certain temperature for normal heating?

PR #94, opened 2021-10-29
Improve documentation

arallsopp:
I've updated the basic documentation in readme.md to include details of functionality and usage, based upon the text that was present in HA before the integration was stripped, to satisfy Issue #65.

I've then reworked sections of it to reflect the change from scan to polling (PR #30).

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,35 @@ To use the component, you will need an application to generate a Tesla refresh t
 8. In the HA UI go to "Configuration" -> "Integrations" click "+" and search for "Tesla Custom Integration".
 
 <!---->
+## Usage
+The `Tesla` integration offers integration with the [Tesla](https://auth.tesla.com/login) cloud service and provides presence detection as well as sensors such as charger state and temperature.
+
+
+This integration provides the following platforms:
+
+- Binary sensors - such as update available, parking, and charger connection.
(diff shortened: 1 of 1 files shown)


arallsopp:
Hey @alandtse,

I realise it was your intention that this would be added to the wiki, but putting it on the readme gives people a chance to view it within HACS. As the documentation expands, we can link the wiki to more detailed pages if required.

Hope that's ok?

alandtse:
Actually, HACS reads from info.md if that's the reasoning. I'm actually indifferent to wiki vs docs. Wiki isn't gated by PRs so is why I suggest that. My hope is having less friction will encourage participation.

arallsopp:
I get you. It makes sense to use the wiki if you feel people are more likely to add to it there.

Re info.md, apologies. I'm not too familiar with creating integrations. When I follow the 'documentation' link in HACS, it just takes me to https://github.com/alandtse/tesla, hence my suggesting adding it to that page. 

Happy to help, and can put it anywhere you see fit, to be honest. You do great work!

alandtse:
Oh that's right. But on the page where you choose the integration in HACS, it reads from info. It's fine in readme because they're likely not looking at the HACS page anymore. I appreciate your help!

arallsopp:
Glad to be able to assist. Thanks for the merge.

PR #110, opened 2021-12-05
Expose charge_current_request_max attribute

danielp370:
Hi,

This PR exposes the charge_current_request_max variable now available in teslajsonpy.  

Tested in HA and it presents the expected value. Ran black over changed files :)

Thanks


alandtse:
It's a bit confusing but the `manifest.json` in the custom_component directory also needs to be updated. That is what HA will read.

The pyproject.yaml is just for development.

danielp370:
First time using poetry - Do I need to push the poetry.lock file as well?

alandtse:
> Do I need to push the poetry.lock file as well?

Yes.

alandtse:
Thanks. There's another feature update in the queue so I'll leave this in dev for a little to see if I can batch them.

PR #111, opened 2021-12-05
Add Horn and Flash Light

rafal83:
Need request on teslajsonpy project

### DIFF
--- a/custom_components/tesla_custom/button.py
+++ b/custom_components/tesla_custom/button.py
@@ -49,4 +49,4 @@ class FlashLights(TeslaDevice, ButtonEntity):
     async def async_press(self, **kwargs):
         """Send the command."""
         _LOGGER.debug("Flash lights: %s", self.name)
-        await self.tesla_device.flash_light()
+        await self.tesla_device.flash_lights()
--- a/custom_components/tesla_custom/button.py
+++ b/custom_components/tesla_custom/button.py
@@ -2,12 +2,16 @@
 from custom_components.tesla_custom.const import ICONS
 import logging
 
-from homeassistant.components.button import ButtonEntity
+_LOGGER = logging.getLogger(__name__)
+
+try:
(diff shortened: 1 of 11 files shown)


alandtse:
This will be a feature bump so it'd be 1.4.0 at a minimum. You can wait on setting that until after we've merged below though.

alandtse:
Ok, I pushed your teslajsonpy changes so it should hit pypi soon. You'll need to resolve the manifest.json conflict, update the pyproject version, and generate a new poetry.lock by running `poetry lock`. See #110 for an example.

custom_components/tesla_custom/button.py
```diff
@@ -0,0 +1,52 @@
+"""Support for Tesla charger buttons."""
+from custom_components.tesla_custom.const import ICONS
+import logging
+
+from homeassistant.components.button import ButtonEntity
```
alandtse, line 5: Let's put a try/catch for an Import error on this since this is a new thing. Please test with a pre 2012.x.x version of HA to ensure we fail gracefully in that case.

custom_components/tesla_custom/switch.py
```diff
@@ -21,7 +21,7 @@ async def async_setup_entry(hass, config_entry, async_add_entities):
         elif device.type == "maxrange switch":
             entities.append(RangeSwitch(device, coordinator))
         elif device.type == "sentry mode switch":
-            entities.append(SentryModeSwitch(device, coordinator))
+            entities.append(SentryModeSwitch(device, coordinator))
```
alandtse, line 24: Please make sure you've run `black`.

alandtse (changes requested):
Thanks. Can you handle these two items please?

alandtse:
Thanks. I just tested and will handle in #114. I believe this should close automatically once I merge it in.

alandtse:
This is available in `dev` until 2012.12 officially releases.

PR #123, opened 2021-12-23
add chromium-tesla-token-generator to list

DoloresHA:
I just had great success using https://github.com/DoctorMcKay/chromium-tesla-token-generator to generate my refresh token. It also made it trivial to verify that my token was not being transmitted to a third party, which as a paranoid delusional I loved. :)

Thank you for this awesome integration!


alandtse:
Thanks. Do you mind adding it to info.md too? That's the in-HACS instructions.

PR #125, opened 2021-12-28
Filter VINs

bassrock:
This adds in support to only add specific Teslas (or none) to homeassistant if you have multiple on your account. For my particular setup I have 2 energy sites and a tesla on my account. However I did not want the tesla added to my homeassistant and getting polled.

## Implementation Details
* I tried to add new tests, but they were failing locally with an error about not finding the tesla integration. I tried to solve this but could not find much information online about how to. Would appreciate any thoughts. 
* The default value for including all teslas is to use a space, this is because the homeassistant options flow will not pass through an empty value to the options flow. So if you add in a list of vins, then decide to remove them, the options flow will not signify that the vin value changed. So you have to use a space to signify this.   Ideally including all vins should be a checkbox and if not checked the text box appears, but I could not find out how to do this in the homeassistant documentation.

## Upgrades 
I did have to upgrade the teslajsonpy integration and the homeassistant integration because of two new features:
* Teslajsonpy - https://github.com/zabuldon/teslajsonpy/pull/263
* Homeassistant - https://github.com/home-assistant/core/pull/49912

Marking this as a draft because of the tests and some uncertainty. I will note that I am using this version of the integration as a daily driver just fine.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,11 +7,11 @@ license = "Apache-2.0"
 
 [tool.poetry.dependencies]
 python = "^3.9"
-teslajsonpy = "^1.4.1"
+teslajsonpy = "^1.4.2"
 
 [tool.poetry.dev-dependencies]
-homeassistant = "^2021.3.4"
-pytest-homeassistant-custom-component = "~=0.3.1"
+homeassistant = "=2021.6.0b0"
+pytest-homeassistant-custom-component = "=0.4.0"
 bandit = "^1.7.0"
 black = {version = ">=21.12b0", allow-prereleases = true}
 mypy = ">=0.812"
--- a/custom_components/tesla_custom/__init__.py
+++ b/custom_components/tesla_custom/__init__.py
@@ -28,9 +28,11 @@ from .config_flow import CannotConnect, InvalidAuth, validate_input
 from .const import (
     CONF_EXPIRATION,
     CONF_WAKE_ON_START,
+    CONF_VINS_TO_EXCLUDE,
     DATA_LISTENER,
     DEFAULT_SCAN_INTERVAL,
     DEFAULT_WAKE_ON_START,
+    DEFAULT_VINS_TO_EXCLUDE,
     DOMAIN,
     MIN_SCAN_INTERVAL,
     PLATFORMS,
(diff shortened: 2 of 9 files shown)


alandtse:
What's the error you're seeing on local tests? Hard to provide advice without more details. Also, I'm not enthused by requiring a space for all as it will generate issues. HA hasn't implemented select lists yet?

Olen:
I also get the pytest-errors about missing integration:

ERROR    homeassistant.config_entries:config_entries.py:711 Cannot find integration tesla_custom

It happens on 7 of 10 tests in tests/test_config_flow.py and I guess the problem is this:

```
    with patch(
        "custom_components.tesla_custom.config_flow.TeslaAPI.connect",
        return_value={
            "refresh_token": TEST_TOKEN,
            CONF_ACCESS_TOKEN: TEST_ACCESS_TOKEN,
            CONF_EXPIRATION: TEST_VALID_EXPIRATION,
        },
    ):
```

The 3 test_option_flow... tests completes successfully.

Olen:
And I think you can use a multi_select in config_flow (however, I have not tested it).

It should be something like:
```
vol.Optional(
    CONF_VINS_TO_INCLUDE, default=DEFAULT_VINS_TO_INCLUDE
): cv.multi_select(ALL_VINS),
```
(Not tested and verified against your code)

pyproject.toml
```diff
@@ -7,11 +7,11 @@ license = "Apache-2.0"
...
 [tool.poetry.dependencies]
 python = "^3.9"
-teslajsonpy = "^1.4.1"
+teslajsonpy = "^1.4.2"
 
 [tool.poetry.dev-dependencies]
-homeassistant = "^2021.3.4"
-pytest-homeassistant-custom-component = "~=0.3.1"
+homeassistant = "=2021.6.0b0"
+pytest-homeassistant-custom-component = "=0.4.0"
```
alandtse, line 14: I wonder if this change is what is breaking pytest.

bassrock: LIkely. I had to update to at least that version of homeassistant though for this function https://github.com/home-assistant/core/pull/49912

alandtse: BTW, this was confirmed to be the cause. https://github.com/alandtse/tesla/issues/170

alandtse: As a note, I figured out the fix if you ever want to pick this up again.

bassrock: @alandtse I actually closed this because it looks like this was kinda implemented in the latest version with the Poll switch.

alandtse:
Oh. I didn't realize you didn't know about the polling switches. Those have been available for a long time. I thought you wanted some way to filter during the device discovery.

PR #127, opened 2022-01-07
Set polling policy in configuration

Olen:
This is a follow up of https://github.com/zabuldon/teslajsonpy/pull/167

Polling Policy can now be set when configuring the integration.

I also added quite extensive documentation on how polling works and what the different knobs will influence..


Olen:
Sorry about the failing tests.  I am unable to run the tests locally as they fail due to an error I can't understand.

`ERROR    homeassistant.config_entries:config_entries.py:711 Cannot find integration tesla_custom`

I believe it is due to these lines:

```
    with patch(
        "custom_components.tesla_custom.config_flow.TeslaAPI.connect",
        return_value={
            "refresh_token": TEST_TOKEN,
            CONF_ACCESS_TOKEN: TEST_ACCESS_TOKEN,
            CONF_EXPIRATION: TEST_VALID_EXPIRATION,
        },
    ):
 ```
 
 I don't really know how to fix these.

**Comment by Olen** [comment]
I have been running the controller form the PR for the helper functions in my installation for a few days now, and the car has been both connected and disconnected, sleeping, driving, online and offline and everything seems to work.
So I'll just activate the PR.

**Comment by alandtse** [comment]
Ok, you'll need a release to merge it. I'll prepare it.

File: custom_components/tesla_custom/manifest.json [hunk 1]
```diff
@@ -4,7 +4,7 @@
   "config_flow": true,
   "documentation": "https://github.com/alandtse/tesla",
   "issue_tracker": "https://github.com/alandtse/tesla/issues",
-  "requirements": ["teslajsonpy==1.4.1"],
+  "requirements": ["teslajsonpy==1.5.0"],
```
alandtse, line 7: We'll also need to update pyproject.toml and run a poetry update.

Olen: Forgot to push the new pyproject.toml

How do I run poetry update in github?

alandtse: It needs to be done at the shell.

Olen:
The problem I have is that I can't commit from my local repo.

If I run `poetry run pytest` the tests validate.
But in .pre-commit-config.yaml it want to run pytest directly, which fails with the mentined error:

```
---------------------------- Captured stderr setup -----------------------------
DEBUG:asyncio:Using selector: EpollSelector                                                              
------------------------------ Captured log setup ------------------------------
DEBUG    asyncio:selector_events.py:59 Using selector: EpollSelector
----------------------------- Captured stderr call -----------------------------
ERROR:homeassistant.config_entries:Cannot find integration tesla_custom
------------------------------ Captured log call -------------------------------
ERROR    homeassistant.config_entries:config_entries.py:678 Cannot find integration tesla_custom
```

And since you explicitly added .pre-commit-config.yaml to the repo, I don't want to mess with it in case it breaks something else...

So I need to manually move my local changes into github, which obviously is not the best way to do it...

But if you could pull pytest out of pre-commit or add  .pre-commit-config.yaml to gitignore, I could run the tests locally before commit as well.

alandtse:
Go ahead and comment out pytest in precommit then. Not sure why it's failing.

PR #128, opened 2022-01-07
Polling interval service

Olen:
This PR allows you to set the polling_interval using a service call.

This is useful if you want to enable more frequent polling in certain locations or certain times of the day, or if you want to limit polling (without disabling it completely) when SOC get low, or in a range of other cases.


alandtse:
Should the polling document be in the wiki vs the repo?

Olen:
Maybe the wiki is a better place?
It does not seem like wiki is enabled for this repo, though...

alandtse:
Wiki was on but limited to collaborators. It's public now.

custom_components/tesla_custom/services.yaml
```diff
@@ -29,3 +29,18 @@ api:
       required: true
       selector:
         object:
+
+polling_interval:
+  description: Set polling interval for updating fresh data from an awake car
+  fields:
+    scan_interval:
+      description: Number of seconds between each poll.  See https://github.com/alandtse/tesla/POLLING_POLICY.md for possible impacts
```
alandtse, line 37: Let's fix the links so we can merge this in. I think this should be ready right?
https://github.com/alandtse/tesla/wiki/Polling-policy

Olen:
This one is ready.  Updated the URL now.

alandtse:
Are you sure this doesn't need a new teslajsonpy?

Olen:
It should not need any changes to teslajsonpy.

I just copied the code for changing `update_inteval`  from here:
https://github.com/alandtse/tesla/blob/33663fb9942d7d6f0ef4a119446132527a090e10/custom_components/tesla_custom/__init__.py#L234

PR #135, opened 2022-01-17
Document polling policy

Olen:
Adding info about the Polling Policy options to README.

`prettier`also auto-cleaned up some other lines in the README .

Just ignore the commit called 'Update test_config_flow.py' - it was already in dev so it does not change anything.

### DIFF
--- a/tests/test_config_flow.py
+++ b/tests/test_config_flow.py
@@ -225,10 +225,18 @@ async def test_option_flow(hass):
 
     result = await hass.config_entries.options.async_configure(
         result["flow_id"],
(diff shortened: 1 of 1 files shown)


Olen:
Not sure how to update the other translation files.

alandtse:
Don't worry about the translation files. That is manual at this time. I've been too lazy to set up a lokalize account/workflow like I have in my alexa_media_player custom component.

Olen:
The only reason I mentioned the translation files is that this is where the strings are, so I updated the english and norwegian.
Hopefully, everyone else will get the english text(?)

alandtse:
Nope. I think missing keys are just blank. Hopefully other speakers notice and submit a PR.

PR #148, opened 2022-02-16
Set DataCoordinator update interval to 10 seconds

ehendrix23:
Currently the DataCoordinator is set to call the update for the integration every minute, this change is to change it to 10 seconds instead.
There are many integrations within HASS that are called every 10 seconds. 

With the polling interval, the API calls to Tesla are not increased unless one sets the polling interval to less than 1 minute. When the update call is made but the polling interval is not expired it would not call the Tesla API.
Further, nothing will be recorded in HASS either as long as there is no state change. So this change will not increase for example DB size or so for HASS. It would just add a bit off overhead which is very minimal.

Advantage of this change is that during certain events (i.e. driving) the polling interval can be set to lower then 1 minute and one can then get more up-to-date data.
Note, TeslaFi for example polls every 10 seconds.


alandtse:
I'm not sure I think this should be a default settings this low. Is there a reason you can't just manually adjust this at your end if you want it this low?

alandtse:
@Olen thoughts on this or perhaps we open the minimum as an option?

alandtse:
Hmm. I just realized that we only use this to clamp the minimum value. Perhaps it's worth allowing it. To be honest, if people wanted this frequency, the correct answer is [websockets](https://github.com/zabuldon/teslajsonpy/issues/275). That was providing 4 updates per seconds while driving if I recall.

ehendrix23:
> websockets

Based on comments I saw following that it seems that Tesla only allows 1 client for streaming and it would thus break the app (or the integration broken if the app takes the stream).

FYI, if the state does not change then nothing is added in HASS irrespective on number of times polled. So this would not add any more records in HASS unless the state has changed.

alandtse:
The concern is polling the Tesla servers so much that they decide to ban third party apps. We're already not officially endorsed. The fact Teslamate caused a worldwide reset of tokens is already not a good sign.

ehendrix23:
> The concern is polling the Tesla servers so much that they decide to ban third party apps. We're already not officially endorsed. The fact Teslamate caused a worldwide reset of tokens is already not a good sign.

That’s why I put it for 10 seconds and not less as TeslaFi polls every 10 seconds as well.

Olen:
I have also used 10 seconds in a personally developed app I used for a long time without any problem. However, I do understand Alans concern. I don't know anything about the size of the user base of TeslaFi, Teslamate and Tesla in HA, so I don't know of changing the interval to 10 seonds here will even be noticeable on Teslas servers.

With that said, I think it is ok to change the minimum to 10 seconds, and use the new service call to adjust the actual polling interval dynamically. 

Also, I have not looked at the code of the official Tesla APP for many years, but from watching its behaviour, I believe it uses websockets as soon as the car wakes up.  If it is correct (and I have not verified this myself) that the websocket API only accepts one client at a time, this is a showstopper for using websockets in HA as long as users might want to use the offical APP in addition to HA.  A simple test here would be to open the APP on two or more phones at the same time, and see if they both get live data from a car that is driving or charging.

alandtse:
Assuming all downloads at the latest are active we're at 2.7K users. Probably a drop in the bucket. Since this is just a clamp, I'm ok with it.

PR #149, opened 2022-02-21
Update polling interval service to set by vin

ehendrix23:
This PR is to update existing polling_interval service to allow polling interval to be set by vin as well.


custom_components/tesla_custom/services.py
```diff
@@ -38,7 +39,8 @@
 SCAN_INTERVAL_SCHEMA = vol.Schema(
     {
         vol.Optional(CONF_EMAIL): vol.All(cv.string, vol.Length(min=1)),
-        vol.Required(CONF_SCAN_INTERVAL, default=DEFAULT_SCAN_INTERVAL): vol.All(vol.Coerce(int), vol.Range(min=0, max=3600)),
+        vol.Optional(ATTR_VIN): vol.All(cv.string, vol.Length(min=1)),
+        vol.Required(CONF_SCAN_INTERVAL, default=DEFAULT_SCAN_INTERVAL): vol.All(vol.Coerce(int), vol.Range(min=-1, max=3600)),
```
alandtse, line 43: What is the impact of -1? Also 0? I realize you didn't add 0 but I just realized we had these.

ehendrix23: -1 would result in it being set back to default update_interval.

There is the integration update_interval, that is what is being used for all VINs unless a VIN has a specific update_interval provided.

- integration update_interval is DEFAULT_SCAN_INTERVAL upon start.
- integration update_interval is changed to new value when calling the service without a VIN
- update_interval for a specific VIN is changed when providing a VIN.
- If update_interval for a specific VIN is provided with value -1 then the VIN specific update_interval is removed and thus revering back to integration update_interval.

Example, 2 VINS: VIN_A and VIN_B
  update_interval = 660

1. call polling interval with scan_interval = 330 => update_interval is now 330. Both VIN_A and VIN_B use this update_interval
2. call polling interval with scan_interval = 10 for VIN_B => update_interval for VIN_B is now 10. VIN_A is still at 330 but  VIN_B is now at 10
3. call polling interval with scan_interval = 180 => update_interval is now 180. VIN_A will be at 180 but  VIN_B is still at 10 as it has a VIN specific update interval
4. call polling interval with scan_interval = -1 for VIN_B => update_interval for VIN_B is back to default which is 180, same as VIN_A
5. call polling interval with scan_interval = 660 => update_interval is now 660. Both VIN_A and VIN_B use this update_interval
 
Could make scan_interval optional and when not provided would be same as -1.
If providing 0 then update_interval will be set to 0, in the end it would mean that the information would be polled every. 10 seconds as that is then how often the update runs from the integration.

alandtse: Ok, let's update the wiki with the explanation.

PR #155, opened 2022-03-01
Add a note in the docs about how to turn on climate via a scene

joestump:
I had a heck of a time running down how to turn on my Tesla's climate via a scene; looks like [I'm not alone](https://community.home-assistant.io/t/help-with-climate-automation-tesla/39644). Here's the scene that finally worked for me:

```yaml
- name: I'm Getting Ready to Leave
  icon: "mdi:car"
  entities:
    climate.tesla_model_y_hvac_climate_system:
      state: heat_cool
```

Based on the docs I was expecting to need to set `state` to `on` and `hvac_mode` to `heat_cool`.


alandtse:
Thanks!

PR #168, opened 2022-03-24
feat: add trigger homelink button (disabled by default)

carleeno:
This PR adds a 'trigger homelink" button to the Tesla integration, which is disabled by default.

It works by taking the current location of the vehicle to feed into the required lat/lon params for the API call, which would normally be the user phone's location when called by the official Tesla app. For users this means it's possible to remotely open/close the homelink device (please always monitor surroundings for safety).

The button will be disabled by default in HA, and would need to manually be enabled by the the user if they wish to utilize it.

If enabled, it would still show as unavailable in these cases:

- if the vehicle does not have the homelink option purchased/installed.
- if the vehicle has no homelink devices programmed.
- if the vehicle is unreachable (as is normal behavior for all entities)

It will, however, show available in all other cases, even if the vehicle is reportedly too far from the homelink device. This is to ensure long polling times wouldn't prevent users from using the button shortly after driving within range of the homelink device.
Pressing the button while the vehicle is actually out of range of the homelink device would result in an error logged in HA (along with any other errors the API request might return).

Tested by:

- creating and running unit tests in teslajsonpy
- running a dev instance of HA with a homelink-equipped Tesla:
  - tested error logging of expected errors (too far from device, etc.)
  - tested real-world ability to open/close a programmed homelink device
  - tested button availability when vehicle is offline, or no homelink devices added.


carleeno:
A decision was made to also make this entity disabled by default, so this will depend on #172 as well

carleeno:
**IMPORTANT:** this must not land without #172

PR #172, opened 2022-03-25
Add ability for sensors to be disabled by default

ehendrix23:
Add the ability for sensors to be disabled by default. 

This is to allow for the vehicle data sensors to be disabled in HASS by default.


alandtse:
Note the release is 2.0.0.

PR #187, opened 2022-04-09
Add Heated Seats support via the Select entity

megabytemb:
# Add Heated Seats support via the Select entity

Limitations:
1. The Heated Seat data relies on the Climate Data from Tesla. This only returns a value while the climate is on. This means that the select entities will stay in their previous state until the climate is turned on in the car. The selects update as expected while Climate is turned on.
2. Heated Seats will only change while climate is enabled. This means that if you want to turn on a heated seat, you need to first turn on the climate control, then change the select to the desired level. This is a limitation of the Tesla API.

I thought about automating turning on the climate if the user sets an option and climate isn't on, but automatically turning on the climate could have unintended consequences.

### DIFF
--- a/custom_components/tesla_custom/const.py
+++ b/custom_components/tesla_custom/const.py
@@ -18,6 +18,7 @@ PLATFORMS = [
     "device_tracker",
     "switch",
     "button",
+    "select",
 ]
 
 ICONS = {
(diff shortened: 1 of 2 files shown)


alandtse:
I assume this is ready to integrate right?

megabytemb:
Same issue as heated steats - need to make it match the app.

If you like, i can close this PR, and roll the changes into the seats PR as a more overall climate control update.

alandtse:
No need to close it. I just noticed the commit message so thought you may have done it.

megabytemb:
Closing this as the code has been rolled into the larger climate controls pr #188

PR #188, opened 2022-04-09
Add Support for Heated Steering Wheel and Seats

megabytemb:
# Add Heated Steering Wheel Switch

Allows users to toggle the Heated Steering Wheel from Home Assistant.

This has a silent restriction where it only works when the Climate is Enabled within a car (Similar to the Heated Seats).

If climate is not enabled, then the switch silently fails.

### DIFF
--- a/custom_components/tesla_custom/climate.py
+++ b/custom_components/tesla_custom/climate.py
@@ -20,7 +20,6 @@ from .helpers import get_device
 _LOGGER = logging.getLogger(__name__)
 
 SUPPORT_HVAC = [HVAC_MODE_HEAT_COOL, HVAC_MODE_OFF]
-import pprint
 
 
 async def async_setup_entry(hass, config_entry, async_add_entities):
--- a/custom_components/tesla_custom/climate.py
+++ b/custom_components/tesla_custom/climate.py
@@ -15,6 +15,7 @@ from teslajsonpy.exceptions import UnknownPresetMode
 
 from . import DOMAIN as TESLA_DOMAIN
 from .tesla_device import TeslaDevice
+from .helpers import get_device
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -97,6 +98,7 @@ class TeslaThermostat(TeslaDevice, ClimateEntity):
             await self.tesla_device.set_status(False)
         elif hvac_mode == HVAC_MODE_HEAT_COOL:
             await self.tesla_device.set_status(True)
+        await self.update_climate_related_devices()
         self.async_write_ha_state()
 
     @TeslaDevice.Decorators.check_for_reauth
@@ -124,3 +126,38 @@ class TeslaThermostat(TeslaDevice, ClimateEntity):
         Requires SUPPORT_PRESET_MODE.
         """
         return self.tesla_device.preset_modes
+
+    async def update_climate_related_devices(self):
+        """Reset the Manual Update time on climate related devices.
+
+        This way, their states are correctly reflected if they are dependant on the Climate state.
+        """
+
+        # This is really gross, and i kinda hate it.
+        # but its the only way i could figure out how to force an update on the underlying device
+        # thats in the teslajsonpy library.
+        # This could be fixed by doing a pr in the underlying library,
+        # but is ok for now.
+
+        # This works by reseting the last update time in the underlying device.
+        # this does not cause an api call, but instead enabled the undering device
+        # to read from the shared climate data cache in the teslajsonpy library.
+
+        climate_devices = [
+            ["switch", "heated steering switch"],
+            ["select", "heated seat left"],
+            ["select", "heated seat right"],
+            ["select", "heated seat rear_left"],
+            ["select", "heated seat rear_center"],
+            ["select", "heated seat rear_right"],
+        ]
+
+        for c_device in climate_devices:
+            device = await get_device(
+                self.hass, self.config_entry_id, c_device[0], c_device[1]
+            )
+            if device is not None:
+                class_name = device.__class__.__name__
+                attr_str = f"_{class_name}__manual_update_time"
+                setattr(device, attr_str, 0)
+                await device.async_update(force=True, wake_if_asleep=False)
--- a/custom_components/tesla_custom/climate.py
+++ b/custom_components/tesla_custom/climate.py
@@ -20,6 +20,7 @@ from .helpers import get_device
 _LOGGER = logging.getLogger(__name__)
 
 SUPPORT_HVAC = [HVAC_MODE_HEAT_COOL, HVAC_MODE_OFF]
+import pprint
 
 
 async def async_setup_entry(hass, config_entry, async_add_entities):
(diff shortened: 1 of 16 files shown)


alandtse:
Thank you. How does the app handle attempts to change the heaters when the climate is off? Does it automatically enable climate? I think we should try to match it.

megabytemb:
The App automatically turns on the climate.

alandtse:
Yes. Please try to match the app behavior as a default.

alandtse:
Just to make sure, this is just waiting for you to let me know it matches the app's behavior by enabling the climate. Are you planning to do that?

megabytemb:
Yeah sorry, i've been snowed under with other stuff at the moment.

Its in my plate to get this PR updated so it matches the app.

shocklateboy92:
Okay, I tried to be patient, but it didn't work.
I'll take a crack at finishing this tonight if @megabytemb  doesn't have time.

@alandtse would you like me to:
- dig into the underlying `tesla_device` and directly try turning the climate on?
- try getting an instance of the `TeslaThermostat` instance and calling a method on it to turn on?
- do the programmatic equivalent of a home assistant "service call"?

megabytemb:
Oh i'm sorry! i didn't know anyone else was waiting on this apart from me!!

i was going to work on it today anyway, so stand by

alandtse:
I'm indifferent to the implementation. Most correct would probably be the tesla_device that comes from teslajsonpy. There's a set of HA specific stuff down in the API library.

shocklateboy92:
> Oh i'm sorry! i didn't know anyone else was waiting on this apart from me!!
> 
> i was going to work on it today anyway, so stand by

No worries man! I was trying to be facetious. 
This is an open source project you're contributing to out of the goodness of your heart - you don't owe anyone anything. 😊

It's awesome that you have time today. If not, I'll do it tomorrow.

megabytemb:
Ok, so I've put some new code in here to get the climate device, and use that to enable HVAC - it will wait for up to 30 seconds before it gives up.

But I've came up against an interesting bug. Climate Data doesn't appear to be shared across all devices. So if I have the HVAC and heater steering wheel on, then turn off the HVAC in Home Assistant, it can take awhile for the heated steering wheel to be reflected (as an off state).

Any ideas on how i could get a shared "climate" state? The climate device doesn't store the raw climate data in the class anywhere, so i can't access it.

alandtse:
Hmmm. You may have to do some message passing so other entities are aware of linked changes. I don't think there's a specific mechanism otherwise to share state between entities.

custom_components/tesla_custom/switch.py
```diff
@@ -49,6 +54,32 @@ def is_on(self):
...
+
+        async with async_timeout.timeout(30):
+            while True:
+                hvac_mode = climate_device.is_hvac_enabled()
+
+                if hvac_mode is True:
+                    _LOGGER.debug("HVAC Enabled")
+                    return True
+                else:
+                    _LOGGER.info("Enabing Climate to activate Heated Steering Wheel")
```
shocklateboy92, line 76: Does this mean it will repeatedly send the climate on command every second for up to 30s?
Wouldn't you rather send it once and wait? Or is it sometimes flaky, and the 30 retries be potentially useful?

alandtse: Yes, this is probably not the best way to implement because it will spam the api. Thinking about this more, it's probably much cleaner to have a single api command that sends the hvac start command, checks ever few seconds to see if it's been enabled, and then issues the second command.

alandtse: This really feels like the wakeup decorator. Instead of waking up the car, it instead is enabling the hvac.

megabytemb:
Ok, did some more work on this.

There's some code in there that i'm kinda not proud of - it feels dirty, but as explained in code comments, its the only way I could force other devices to update based on the climate device. I'm fully aware that setting a double underscored class variable externally to the class is really not good python code. haha. 

This really highlighted the need to remove any home assistant stuff from the underlying library, so we can have all the code that influences home assistant logic in a single library. 

I'd be really interested in knowing your thoughts here.

custom_components/tesla_custom/climate.py
```diff
@@ -124,3 +126,38 @@ def preset_modes(self) -> list[str] | None:
...
+    async def update_climate_related_devices(self):
+        """Reset the Manual Update time on climate related devices.
+
+        This way, their states are correctly reflected if they are dependant on the Climate state.
+        """
+
+        # This is really gross, and i kinda hate it.
+        # but its the only way i could figure out how to force an update on the underlying device
+        # thats in the teslajsonpy library.
+        # This could be fixed by doing a pr in the underlying library,
```
shocklateboy92, line 139: By adding a public method to allow resetting the manual update time? 
Perhaps it's worth filing an issue on the `teslajsonpy` repo, so this doesn't get forgotten.

shocklateboy92 (approved):

custom_components/tesla_custom/helpers.py
```diff
@@ -0,0 +1,73 @@
...
+                _LOGGER.debug("HVAC Enabled")
+                return True
+            else:
+                _LOGGER.info("Enabing Climate to activate Heated Steering Wheel")
+
+                # The below is a blocking funtion (it waits for a reponse from the API).
+                # So it could eat into our timeout, and this is fine.
+                # We'll try to turn the set the status, and check again
+                try:
+                    await climate_device.set_status(True)
```
alandtse, line 62: How often on your system does this hit the Tesla endpoint? The issue I'm concerned about is that each api call has a wake call which includes 5 separate retries. Since this loop in theory can hit 15 times and if there's a wakeup that quickly goes to 75 api hits. Are we slamming the endpoint unnecessarily?

custom_components/tesla_custom/climate.py
alandtse: Yes, let's open an issue. Basically what you're saying we need is a way to force a device to reread its info from cache.

I'm ok with the current approach for a 1.0 type release, but it makes sense to resolve it more cleanly in the future.

megabytemb:
Ok, combining your feedback and more testing, I've more tightly controlled when we go to the Tesla API for an update, and forcing other devices to read from cache.

We're actually a little behind on feature parody with tesla's app, and i this this is due to the need to update 2 libraries to implement a new feature.
Want me to open up a separate issue on this library so we can discuss the idea of moving all home assistant logic into this library, and purely using teslajsonpy as the transport layer?
That way we can implement things like a shared car state across entities, and develop with more speed. There's a bunch in the API that we should probably look to implement for feature completeness.

alandtse:
Great. This looks better. Can you rebase off dev so I can then merge this in please?

We already have that issue to do the separation. It's just been a lack of motivation to be honest. https://github.com/zabuldon/teslajsonpy/issues/24

custom_components/tesla_custom/climate.py
```diff
@@ -20,6 +20,7 @@
 _LOGGER = logging.getLogger(__name__)
 
 SUPPORT_HVAC = [HVAC_MODE_HEAT_COOL, HVAC_MODE_OFF]
+import pprint
```
shocklateboy92, line 23: nit: I think you're adding an import but not really using it

megabytemb: fixed

megabytemb:
I think we're in a good place now for a merge into dev.

alandtse:
FYI @megabytemb  you need to use a dict.get(). See https://github.com/alandtse/tesla/issues/199

megabytemb:
Created
https://github.com/zabuldon/teslajsonpy/pull/308
in the teslajsonpy library, where the code is located.

PR #205, opened 2022-04-26
Enable heated seat and steering wheel based on capability of vehicle

ehendrix23:
This PR will enable heated seats and heated steering wheel entities based on the capability of the vehicle.
This allows those entities initially to be created as disabled, but first time climate data is received those that are available are then enabled.
Enabling is only checked once per startup once data has been received. Once an entity is enabled is will stay enabled.
An entity that was disabled by the user will not be re-enabled again.

If approved then I plan to update teslajsonpy to make all heated seats and the heated steering wheel switch as disabled by default. 
Note, that anyone who already has those entities now will not see any change since they are then already enabled. This will only be seen when newly adding the component.

### DIFF
--- a/custom_components/tesla_custom/manifest.json
+++ b/custom_components/tesla_custom/manifest.json
@@ -22,5 +22,5 @@
     }
   ],
   "iot_class": "cloud_polling",
-  "version": "2.1.2"
+  "version": "2.1.1"
 }
--- a/custom_components/tesla_custom/manifest.json
+++ b/custom_components/tesla_custom/manifest.json
@@ -4,7 +4,7 @@
   "config_flow": true,
   "documentation": "https://github.com/alandtse/tesla/wiki",
   "issue_tracker": "https://github.com/alandtse/tesla/issues",
-  "requirements": ["teslajsonpy==2.0.3"],
+  "requirements": ["teslajsonpy==2.1.0"],
   "codeowners": ["@alandtse"],
   "dependencies": ["http"],
   "dhcp": [
@@ -22,5 +22,5 @@
     }
   ],
   "iot_class": "cloud_polling",
-  "version": "2.1.1"
+  "version": "2.1.2"
 }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,13 +1,13 @@
 [tool.poetry]
 name = "tesla"
-version = "2.1.1"
+version = "2.1.2"
 description = "A fork of the Home Assistant tesla integration using a oauth proxy to login."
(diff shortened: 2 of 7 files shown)


alandtse:
Just so we avoid deadlock, I'm assuming you need to bump teslajsonpy so this isn't ready for integration. Let me know if I'm wrong. I try to do releases on weekends in case that helps your planning. No rush.

ehendrix23:
> Just so we avoid deadlock, I'm assuming you need to bump teslajsonpy so this isn't ready for integration. Let me know if I'm wrong. I try to do releases on weekends in case that helps your planning. No rush.

Yep, just bumped teslajsonpy version and updated other files for new version of this item unless I missed something.
Also prepped changelog but commit still has to be updated since this hasn't merged yet. Do check make sure it looks good. :-)

alandtse:
Please note that changelogs are autogenerated by the commit message.

ehendrix23:
> Please note that changelogs are autogenerated by the commit message.

Do I need to push a new commit to revert that then?

alandtse:
Yes. Otherwise iti'll just have to entries for 2.1.2

pyproject.toml
```diff
@@ -1,13 +1,13 @@
 [tool.poetry]
 name = "tesla"
-version = "2.1.1"
+version = "2.1.2"
```
alandtse, line 3: Same here.

pyproject.toml
```diff
@@ -1,13 +1,13 @@
...
 description = "A fork of the Home Assistant tesla integration using a oauth proxy to login."
 authors = ["Alan D. Tse <alandtse@gmail.com>"]
 license = "Apache-2.0"
 
 [tool.poetry.dependencies]
 python = "^3.9"
-teslajsonpy = "^2.0.3"
+teslajsonpy = "^2.1.0"
```
alandtse, line 10: This is good to bump.

pyproject.toml
ehendrix23: Reverted

custom_components/tesla_custom/manifest.json
```diff
@@ -22,5 +22,5 @@
     }
   ],
   "iot_class": "cloud_polling",
-  "version": "2.1.1"
+  "version": "2.1.2"
```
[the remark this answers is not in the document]
ehendrix23: One_sap922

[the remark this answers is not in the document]
ehendrix23: Reverted

Issue #207, opened 2022-05-01
Issue when multiple cars are tied to the same tesla account (heated steering and seats)


rako77:
https://github.com/alandtse/tesla/blob/dbd4480a1c4445b610454dbc3ac8e2639bd7611f/custom_components/tesla_custom/helpers.py#L17

In this helper and specifically my issue is with an account with multiple cars fails to turn on the correct climate when enabling the heated steering wheel or heated seats.

The issue appears to be it's returning the first device in the array of devices without checking that the device is the right one, therefor In my home when the heated steering for Car \# 2 is triggered it turns on Car \# 1 climate.

alandtse:
@megabytemb @ehendrix23 can either of you confirm please?

megabytemb:
I wish i had the infrastructure to properly test this. haha
But i can see how that problem would arise.

I don't know exactly how I would fix it right now, But i'll get this done after work (or perhaps at lunch) today.

ehendrix23:
@megabytemb Let me know if you need help. I have 2 cars in my account. Only 1 with heater steering wheel but both have heated seats. :-)

megabytemb:
Created https://github.com/alandtse/tesla/pull/208 which i believe fixes this issue.

PR #208, opened 2022-05-02
Match Devices to the same VIN.

Megabytemb:
When we are updating other devices, we need to ensure we do so on the same device on the same car.

Fixes https://github.com/alandtse/tesla/issues/207

### DIFF
--- a/custom_components/tesla_custom/climate.py
+++ b/custom_components/tesla_custom/climate.py
@@ -157,6 +157,7 @@ class TeslaThermostat(TeslaDevice, ClimateEntity):
         vin = self.tesla_device.vin()
 
         # Get all the climate parameters so we can determine what is supported by this vin.
+        # pylint: disable=protected-access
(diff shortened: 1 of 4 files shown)


rako77:
thanks for this, I did find that the heated steering wheel switch hasn't been updated to include the vin.
https://github.com/alandtse/tesla/blob/dbd4480a1c4445b610454dbc3ac8e2639bd7611f/custom_components/tesla_custom/switch.py#L38

megabytemb:
Good spot.
Updated.

I'm working on a complete re-write of the entity handing in this component- https://github.com/megabytemb/tesla/tree/tesla-rewrite 
I'll do my best to have complete test coverage to prevent issues like missed params like that.

alandtse:
Thanks. I'll merge and release as soon as I get confirmation this fixes the issue. @rako77

rako77:
Thank you, Confirmed working with my home setup, thank you so much.

PR #216, opened 2022-05-09
DRAFT: Tesla rewrite

Megabytemb:
This is a WIP branch for a Home Assistant Rewrite where we move all Home Assistant related code directly into this library.

This will make development of the Component simpler, and allow us to have a quicker development cycle for new features from Tesla.

I've made an effort to keep the same unique ids, so entities behave the same, but there are some extra attribute changes as I removed duplication.

Also, this has introduced some new entities such as the update entity, and number entities to make better use of Home Assistants new entity types.

This is well and truly a breaking change, and using this branch should be done with caution.

Things still do be done:
- Create media_player entity
- write tests for All code.
- Clean up code comments and documentation.
- Clean-up and review services.py
- re-create app-like functionality, where the climate is enabled before heated seats are enabled.


shred86:
Thanks for starting this. I'm a python newb but I have worked on some Home Assistant integrations so I can maybe help. I'm mostly interested in implementing Tesla solar stuff. I'm not sure what the best way to handle both cars and solar systems would be perhaps make `TeslaBaseEntity` a bit more generic and create two classes that inherit it, `TeslaCar` (basically what you already have but just move car specific stuff in there) and `TeslaEnergy` that covers all energy specific stuff (solar systems, solar roofs, power walls)?

Additionally, a lot of the class names could probably be renamed to add `Car` somewhere in their name since there's some commonality with energy devices like batteries. For example, `TeslaBattery` could perhaps be `TeslaCarBattery`?

Anyways, more than happy to help where I can. The energy site stuff should be fairly easy to implement. I've already made a [pull request](https://github.com/zabuldon/teslajsonpy/pull/333) to teslajsonpy to add it so I'm familiar with how the "API" works.

shred86:
So I've started working on this since I have some free time this week. What I've done so far is:
- Move car specific code out of `TeslaBaseEntity` to a newly created `TeslaCarEntity`.
- Created a new `TeslaEnergyEntity` which also inherits from `TeslaBaseEntity`.
- Creating `TeslaEnergySolarPowerSensor`, TeslaEnergyLoadPowerSensor`, `TeslaEnergyGridPowerSensor` and `TeslaEnergyBatteryPowerSensor`
- Changes to teslajsonpy `Controller` which should support both the current integration and re-write. However, I'm assuming we'll probably want to clean up teslajsonpy once this re-write is done.

So far this seems to be working, although that's not saying much 😆 as I'm definitely not a developer by any means, just a part-time hobbyist who has worked with a couple HA integrations and libraries.

@alandtse What's the best way to start submitting PRs? Just make a fork of the fork/branch @Megabytemb created and submit a PR on his branch? I'm just not sure if he will have time to merge them so I was thinking if we can maybe create a branch in your repo, merge his changes, then we could just fork that and all submit changes to there?

custom_components/tesla_custom/update.py
```diff
@@ -0,0 +1,111 @@
...
+        version_str = self.car.state.get("car_version")
+
+        # We will split out the version Hash, purely cause it looks nicer in the UI.
+        if version_str is not None:
+            version_str = version_str.split(" ")[0]
+
+        return version_str
+
+    @property
+    def in_progress(self) -> int | bool:
```
shred86, line 81: `int | bool` is throwing an error: `unsupported operand type(s) for |`

custom_components/tesla_custom/update.py
```diff
@@ -0,0 +1,111 @@
...
+        # If its actually installing, we can use the install_perc
+        if update_status == "installing":
+            progress = self.car.state.get("software_update", {}).get("install_perc")
+            return progress
+
+        # Otherwise, we're not updating, so return False
+        return False
+
+    async def async_install(
+        self, version: str | None, backup: bool, **kwargs: Any
```
shred86, line 98: Same comment as above regarding `|`

custom_components/tesla_custom/update.py
alandtse: I believe this is tied to python 3.10. What version of python are you using?

shred86: Using 3.9

shred86:
Ah okay, makes sense.

Yeah, I was looking at TeslaPy but realized it wasn't async... started thinking about maybe we should just help rewrite that to be async but that's probably not an easy task (well beyond my knowledge).

I think teslajsonpy, with a little modification, will still allow us to re-write this integration to do as you mentioned. My thought was once we get that done, we can then clean up teslajsonpy to make it more generic which will probably require some updates to the integration as well. I would like to also setup the config flow in HA so you can use your username and password, and the integration automatically grabs the refresh token (or is there a reason it's not like that already). Anyways, I'll work on what I can and submit a PR here.

alandtse:
Closing as integrated with #250

PR #217, opened 2022-05-26
async_get_registry is deprecated in 2022.6, use async_get instead

jasperslits:
The call to async_get_registry in tesla_device.py gives a deprecation warning in 2022.6b0, this PR changes the async call to async_get. 
This should work for 2022.5.x too but haven't been able to test that.

### DIFF
--- a/custom_components/tesla_custom/tesla_device.py
+++ b/custom_components/tesla_custom/tesla_device.py
@@ -5,7 +5,7 @@ from typing import Any, Optional
 
 from homeassistant.const import ATTR_BATTERY_CHARGING, ATTR_BATTERY_LEVEL
 from homeassistant.core import callback
-from homeassistant.helpers.entity_registry import async_get_registry
+from homeassistant.helpers import entity_registry as er
 from homeassistant.helpers.update_coordinator import CoordinatorEntity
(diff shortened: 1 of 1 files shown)


jasperslits:
I don’t believe the bump is needed. It now just shows a warning but I don’t see anything on the removal. I think it typically takes another 2 releases at least.

alandtse:
Oh, I meant whether you know if the `async_get` call exists in the older version of HA or not. This way we don't break people on older versions because we just enforce a minimum HA version.

jasperslits:
> Oh, I meant whether you know if the `async_get` call exists in the older version of HA or not. This way we don't break people on older versions because we just enforce a minimum HA version.

I couldn't find the original commit, but it's been there for 15+ months: https://github.com/home-assistant/core/blame/dev/homeassistant/helpers/entity_registry.py

alandtse:
Thanks. For future reference, if you can find the commit that brings it in, that will give you a list of all the HA versions that are supported.
https://github.com/home-assistant/core/commit/ed31cc363b3a457ed9249f7cf0386f2ec86e1850

In this case, we're still supported. Thanks for checking and for the PR!

PR #218, opened 2022-05-28
Add support for async_remove_config_entry_device

bdraco:
Enables the Delete device button on the devices page

I had duplicate devices since at some point the identifiers had changed

Screen_Shot_2022-05-28_at_11_22_05


bdraco:
Thanks

alandtse:
No prob. Thanks for the PR. It'll be in dev if you want it now. Otherwise, I need to fix up the HACS errors before I actually do a release.

bdraco:
No rush. I already cleaned up my stale devices, but I figured others would benefit from it

PR #244, opened 2022-08-06
Bump teslajsonpy to 2.4.1

shred86:
Adds grid and home (load) power sensors for energy sites (solar systems).

### DIFF
--- a/custom_components/tesla_custom/manifest.json
+++ b/custom_components/tesla_custom/manifest.json
@@ -4,7 +4,7 @@
   "config_flow": true,
   "documentation": "https://github.com/alandtse/tesla/wiki",
   "issue_tracker": "https://github.com/alandtse/tesla/issues",
-  "requirements": ["teslajsonpy==2.3.0"],
+  "requirements": ["teslajsonpy==2.4.0"],
   "codeowners": ["@alandtse"],
   "dependencies": ["http"],
   "dhcp": [
--- a/custom_components/tesla_custom/manifest.json
+++ b/custom_components/tesla_custom/manifest.json
@@ -4,7 +4,7 @@
   "config_flow": true,
   "documentation": "https://github.com/alandtse/tesla/wiki",
   "issue_tracker": "https://github.com/alandtse/tesla/issues",
-  "requirements": ["teslajsonpy==2.4.0"],
+  "requirements": ["teslajsonpy==2.4.1"],
   "codeowners": ["@alandtse"],
   "dependencies": ["http"],
   "dhcp": [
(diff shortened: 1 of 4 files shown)


alandtse (changes requested):
Thanks. Can you also make sure to update pyproject.toml and do a poetry update instead of modifying poetry.lock directly?

shred86:
Sorry about that - clearly have no idea what I'm doing with poetry. 😆 

So I just realized the only reason the solar sensor is being recreated is because I renamed it from `solar panel` to `solar power` to align the naming convention, `solar power`, `grid power`, `load power`. Instead of creating a breaking change, I can simply just retain the `solar panel` name so the `unique_id` in Home Assistant doesn't change...

shred86:
Just made a [pull request](https://github.com/zabuldon/teslajsonpy/pull/336) over on teslajsonpy to revert the name. Sorry for the hassle and I should have caught this earlier but this will prevent the breaking change.

alandtse:
Thanks for the patience. Just returned from vacation. 2.4.1 is pushed with the reversion.

shred86:
Welcome back! Just updated the PR to include version 2.4.1 for teslajsonpy. I also updated the version of this integration to 2.4.0 in the manifest.json which is what I'm assuming it will be.

alandtse:
You need to run `poetry update`. It should auto update dependencies and find 2.4.1.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "tesla"
-version = "2.3.1"
```
alandtse, line 3: Don't bump the component version. That's done by CI.

alandtse (changes requested):

custom_components/tesla_custom/manifest.json
```diff
@@ -22,5 +22,5 @@
     }
   ],
   "iot_class": "cloud_polling",
-  "version": "2.3.1"
```
alandtse, line 25: Don't change this.

pyproject.toml
shred86: Updated! Sorry about that.

custom_components/tesla_custom/manifest.json
shred86: Updated! Sorry about that.

shred86:
Sorry it wasn't updating to 2.4.1 when doing `poetry update`. It finally did after it was published for 13 minutes... guess I was just being impatient. 😄

alandtse (approved):

PR #263, opened 2022-09-11
fix #233 Car location offset

### LINKED ISSUE #233: Car location offset
Hello Alandtse, thanks for you great integration!

My model3 is in China. The gps info (longitude and latitude) obtained through teslamate's mqtt is accurate, but the gps info obtained through homeassistant's Tesla integration (v2.3.0) will always have a constant offset.

For example:

GPS info from teslamate：

- latitude: 23.127603 longitude: 113.318353 (actual location)

GPS info from Tesla integration v2.3.0:

- latitude: 23.125008 longitude: 113.323774

Thanks again.

Emniroll:
Add LocationConverter when users use https://auth.tesla.cn to login.


alandtse:
Thanks!

PR #288, opened 2022-10-22
Use car distance units & remove solar type from model

### LINKED ISSUE #284: No longer shows in Miles only km
New version 3.x shows everything in Kms and not miles. Why has this changed?

All metrics in Kms rather than miles.

Reinstate miles as option. 

**Additional context**
Was present in 2.x correctly.

shred86:
Fixes #284

### Changes

- Checking the vehicles `gui_distance_units` to determine whether to display miles vs kilometers instead of the user's Unit System setting in Home Assistant.
- Remove `solar_type` from model. The Tesla API recently changed where `solar_type` is no longer included in the `PRODUCT_LIST` endpoint.


alandtse:
Ugh, just saw this with the nonstandard commits. I'll revert and fix.

jaymunro:
solar_type is still present, but it is not at the top level, but under `components`. See my comment in [#354 of teslajsonpy](https://github.com/zabuldon/teslajsonpy/issues/354).

Hence shouldn't the change be the following?
```
-        model = f"{self._energysite.resource_type.title()} {self._energysite.solar_type.replace('_', ' ')}"
+        model = f"{self._energysite.resource_type.title()} {self._energysite.components.solar_type.replace('_', ' ')}""
```

I realise this has already been merged but I have never done a PR and don't know if I have the perms to start one.

Issue #282, opened 2022-10-22
Upgrade 3.0 Tesla Location tracker not work (China Server)


BlueBenson:
## Version of the custom_component
3.0
## Configuration

```yaml
Add your logs here.
```

## Describe the bug

Upgrade 3.0 Tesla Location tracker not work (China Server)
Wendy's TeslaX Location tracker was detected at unknown

## Debug log

```text

Add your logs here.

```

alandtse:
Please enable debug logs for tesla and teslajsonpy and check for any errors.

alandtse:
Please also confirm:
1. You've removed and readded the component.
2. You've used the Chinese auth server to login.
3. Other features work.

BlueBenson:
Yes I confirm. I deleted the old component and reinstalled. I used the china auth. Except the GPS other functions all good. Sent from my iPhoneOn Oct 22, 2022, at 18:51, Alan Tse ***@***.***> wrote:﻿
Please also confirm:

You've removed and readded the component.
You've used the Chinese auth server to login.
Other features work.

goxofy:
same problem

China server(tesla.cn)

BlueBenson:
> Please enable debug logs for tesla and teslajsonpy and check for any errors.

2022-10-22 22:20:13.896 DEBUG (MainThread) [custom_components.tesla_custom] Connected to the Tesla API
2022-10-22 22:20:13.897 DEBUG (MainThread) [custom_components.tesla_custom] Running controller.update()
2022-10-22 22:20:18.614 DEBUG (MainThread) [custom_components.tesla_custom] Finished fetching tesla_custom data in 4.718 seconds (success: True)
2022-10-22 22:20:18.639 DEBUG (MainThread) [custom_components.tesla_custom] Running controller.update()
2022-10-22 22:20:18.640 DEBUG (MainThread) [custom_components.tesla_custom] Finished fetching tesla_custom data in 0.000 seconds (success: True)
2022-10-22 22:20:28.000 DEBUG (MainThread) [custom_components.tesla_custom] Running controller.update()
2022-10-22 22:20:28.000 DEBUG (MainThread) [custom_components.tesla_custom] Finished fetching tesla_custom data in 0.000 seconds (success: True)
2022-10-22 22:20:38.001 DEBUG (MainThread) [custom_components.tesla_custom] Running controller.update()
[...]

PR #289, opened 2022-10-22
fix: use GPS coordinates for users in China

shred86:
Based on what I'm seeing in #282 and #233, it appears the Tesla API is already passing GCJ-02 coordinates for users in China in WGS-84 format. This was also [confirmed](https://github.com/alandtse/tesla/issues/233#issuecomment-1182689492) by a user in China.

I'd like to merge this to dev and have some of our users in China test this out for us just to make sure, but everything I'm seeing suggests we can just use the `longitude` and `latitude` values passed directly from the Tesla API for all users since it appears to in WGS-84 format which is what Home Assistant uses.


alandtse:
Since you'll be helping to maintain, please start using [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) messages. Also, please make sure you always start branches from dev.

Given we need to push out a bunch of bug fixes, my plan will be to just push multiple fixes this weekend without a long dev cycle. Users don't really understand dev so waiting for feedback from there wouldn't iterate very fast.

PR #292, opened 2022-10-22
fix: set sensor device class

### LINKED ISSUE #284: No longer shows in Miles only km

shred86:
#### Change

- Specify device class for sensors to allow users to specify the unit of measurement for each entity.

#### Details

- Reverts a [previous PR](https://github.com/alandtse/tesla/pull/288) but this still fixes #284. Specifically, if a UK user has their Home Assistant Unit System set to metric, they can still set the units for the odometer to miles (one example).

### DIFF
--- a/custom_components/tesla_custom/sensor.py
+++ b/custom_components/tesla_custom/sensor.py
@@ -16,7 +16,6 @@ from homeassistant.const import (
     PERCENTAGE,
     POWER_WATT,
     POWER_KILO_WATT,
-    SPEED_KILOMETERS_PER_HOUR,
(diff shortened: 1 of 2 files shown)


alandtse:
Please note, it's not just the PR title that should follow the convention. It's the actual commit itself. There are plugins that help enforce this like [comittizen](https://github.com/commitizen/cz-cli). I'm squashing these to fix these for you, but once you figure it out you can just start merging.

Issue #234, opened 2022-07-12
Windows vent switch/window control


ilvar:
Since Cabin Overheat Protection requires Tilt/Intrusion Alarm to be **off** and Sentry Mode requires Intrusion Alarm to be **on**, we can't have both Overheat Protection and Sentry Mode. 
 
I think it would be nice to have a Switch or Window exposed to activate Vent when the car is at Home, and it's too hot outside.

I've tried to use HVAC controls for that, but it's less reliable. Also, turning HVAC off when the cabin cools down means that it will be off during the next ride.

InTheDaylight14:
I made a custom cover a while back using the custom tesla API call service. It's not working right now but was nice for a while. 

```
cover:
  - platform: template
    covers:
      tesla_windows:
        device_class: window
        friendly_name: "Tesla Windows"
        open_cover:
          - service: tesla_custom.api
            data:
              command: WINDOW_CONTROL
              parameters:
                path_vars:
                  vehicle_id: 'my_id'
                command: vent
                lat: 0
                long: 0
                wake_if_asleep: true
        close_cover:
          - service: tesla_custom.api
            data:
              command: WINDOW_CONTROL
              parameters:
                path_vars:
                  vehicle_id: 'my_id'
                command: close
                lat: 0
                long: 0
                wake_if_asleep: true
```

miguelcastilho:
> 
How did you add this custom Tesla API to home assistant? As far as I know this integration does not support interaction with the windows.

bkr1969:
> Since Cabin Overheat Protection requires Tilt/Intrusion Alarm to be **off** and Sentry Mode requires Intrusion Alarm to be **on**, we can't have both Overheat Protection and Sentry Mode.

Is this actually true?  I have my car on Sentry mode at work all day and still have cabin overheat protection on (and it activates, and I have a Node Red automation set to vent the windows even before overheat protection kicks in if the weather is nice). Where do you find the info that both can't run simultaneously?

ilvar:
> I have my car on Sentry mode at work all day and still have cabin overheat protection on 

It might be a thing of European software flavor, not sure. Do you have Tilt/Intrusion Alarm on as well?

PR #318, opened 2022-10-31
feat: Add window cover, add covers to readme

InTheDaylight14:
Tested window functionality based on my fork of teslajsonpy in the dev container. I was able to get it working. Addresses https://github.com/alandtse/tesla/issues/234

One note: I had to comment out lines with "CoverEntityFeature" in the dev container for the cover entities to load on setup (I assume that will work with future HA versions?). I left them intact for this PR.


alandtse:
You need to bump teslajsonpy now. In the future link the dependency pr so I know.

InTheDaylight14:
Looks like teslajsonpy 3.0.0 was being installed for the tests. Sorry, still new to this

tests/test_cover.py
```diff
@@ -78,3 +78,26 @@ async def test_trunk(hass: HomeAssistant) -> None:
...
+            {ATTR_ENTITY_ID: "cover.my_model_s_window"},
+            blocking=True,
+        )
+        mock_open_cover.assert_awaited_once()
+
+    with patch("teslajsonpy.car.TeslaCar.close_windows") as mock_close_cover:
+        assert await hass.services.async_call(
+            COVER_DOMAIN,
+            SERVICE_CLOSE_COVER,
+            {ATTR_ENTITY_ID: "cover.my_model_s_windows"},
```
alandtse, line 100: Do you need to remove the s here too?

InTheDaylight14:
I'm not sure why the test is failing. When running them on my local machine I'm getting a different error. All the tests fail with: 
``` pytest_socket.SocketBlockedError: A test tried to use socket.socket.```

Any guidance on what I can do to figure it out?

**Comment by alandtse** [comment]
> When running them on my local machine I'm getting a different error. All the tests fail with:
> ` pytest_socket.SocketBlockedError: A test tried to use socket.socket.`

I haven't looked at the tests, but in theory tests should not actually touch the network but instead should have the API response mocked so no API is actually touched.

**Comment by InTheDaylight14** [comment]
Could this be an issue when installing the dependencies in the tests? 
```
Warning: poetry.lock is not consistent with pyproject.toml. You may be getting improper dependencies. Run `poetry lock [--no-update]` to fix it.
```

I can commit the updated lock file.

alandtse:
Yes, you need to run `poetry update` and commit the lock file whenever you change/update dependencies. However, that won't solve the issue if the mocks are hitting apis.

alandtse:
It's failing because self.is_closed is `true` so it doesn't try to close it again.
https://github.com/alandtse/tesla/pull/318/files#diff-3ebb2677d043645911800a521a02da6fa68181b50949d10428055114372a9b5bR176

You'll notice that other covers don't have this requirement at all.

The issue is you're mocking the underlying teslajsonpy call, which is responsible with updating the car state. Because the call is not happening, HA's state isn't changing.

The way for the test to work is to set the car window to open and then run the service_call. You may want to look at HA cover tests to see the right way to set an entity's state. It's been a while since I looked at it.

InTheDaylight14:
Thanks for the info, I think I'll be able to do it now. I'm going to start from the top if that's okay. This got messy

alandtse:
For reference, the reason for the last change was explained here. https://github.com/alandtse/tesla/pull/338

alandtse:
Is the API smart enough that trying to open or close a window when it's already open or closed isn't a problem for the car? Even though we have a warranty disclaimer, let's not purposely release something that can cause physical damage.

InTheDaylight14:
> Is the API smart enough that trying to open or close a window when it's already open or closed isn't a problem for the car? Even though we have a warranty disclaimer, let's not purposely release something that can cause physical damage.

Are any other changes needed?

Issue #323, opened 2022-11-02
Bring back release charge port option


aalllop:
**Is your feature request related to a problem? Please describe.**
On previous versions it had the option to release the charge port. Now it has disappeared 

**Describe the solution you'd like**
Bring back that entity

jcmallorquin:
Yes, please. Very useful!

Woempia:
Yes please, it was so useful. Really miss it

agehall:
Yes! This was an essential part of my HA integration for my car!

olsenab:
Same for me, the only function i used :)

InTheDaylight14:
Do y'all have the entity name from before? I can try to keep it the same when adding this back.

I assume it was a lock. "lock.my_tesla_charge_port_latch"? @aalllop

agehall:
it was called “lock.<car_name>_charger_door_lock” before.

Great if it can be kept but I don’t really care that much - the important thing is that I can use it. If I have to change a few automations, no problem…

alandtse:
The name will have to change. The breaking change was because of that rename. @shred86 what was the convention you adopted here?

agehall:
Are you sure it was just the name that changed? I thought this was caused by the charge port entity becoming a door instead of a lock...

alandtse:
Yes there were other changes but the name change itself would've been considered breaking because the type (e.g., lock) was removed from the end of entities. Breaking changes means users have to fix stuff as it's not seamless. Since shred wanted to do that anyway, we bundled a bunch.

agehall:
Sure, but a breaking name change is not a big deal IMHO - that just means I have to spend 10 minutes adjusting my integrations. If the entity disappears completely, which was the case here, we are kind of stuck unless one goes down the route of using the custom api service.

shred86:
This was addressed in [this issue](https://github.com/alandtse/tesla/issues/285). The functionality is still there, you just have to call the cover open service.

Issue #285, opened 2022-10-22
Upgrade 3.0 unable to unlock charger door


copacetic82:
Regarding: "Changed trunk, frunk and charger door from lock to cover entities."

In the previous version I used the charger door lock entity to unlock the charge port i.e. to be able to remove the charging cable. With version 3.0 I am unable to do this. I have to unlock the door from the official Tesla app. 
- Opening and closing the charger door works when no cable is connected.
- Unlocking the car using the lock entity does not unlock the charger door/port
- Switching off the charger switch does not unlock the charger door/port

alandtse:
Thanks. @shred86 I haven't tested this, but did we consider the case the charger door could lock? This may just mean having the cover issue an unlock command on open.

ad2022uk:
Yes I can unlock the charger either.
I used to invoke the unlock_charger option before.

shred86:
We're calling the same `CHARGE_PORT_DOOR_OPEN` as before. All we did was essentially change the previous charger door type from a lock entity to a cover entity, but the underlying functionality is identical.

However, I just realized one change in that before, I think you could issue the lock and unlock command regardless of the state the charger door was in. With the cover entity, if the door is open, it doesn't let you select the "open" button in the HA front end. However, you can call the "Cover: Open" service which will work. Just tested this on my Model S.

@copacetic82 If you unlock the car using the lock entity, you can't remove the charger cable? But if you unlock the doors from the official Tesla app, you can? That doesn't make sense to me either. I just tested this on my car and it works. I did notice if I unlock the doors, the charger port ring on the car is still green (latched) but I can remove the charger cable by pressing the button on the charger itself. When I issue the "Cover: Open" command, the charger port ring is white (unlatched) so I can just pull the charger plug out without pushing the button.

copacetic82:
@shred86 I was referring to the charger door ("I have to unlock the door from the official Tesla app."), sorry for the misunderstanding. ;) 

I can confirm that (while charger door open and cable locked) the cover shows "open" in the frontend i.e. "open" button is grey and only "close" button can be pressed. Calling the entity manually as shown below does work (thanks for the tip!). I can build an automation to work around this and unlock the door from the frontend. :)

service: cover.open_cover
data: {}
target:
  entity_id: cover.tesla_model_3_charger_door

PR #334, opened 2022-11-07
feat: Add charge port latch lock, #323

InTheDaylight14:
https://github.com/alandtse/tesla/issues/323

Feature working in the dev container. 

There is no option to close the latch with Tesla's API. The car itself handles that. I could not find a way to hid the "lock" feature so I'm having it throw a message in the logs that it cannot be locked. I'm not sure what is best practice in that situation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ This integration provides the following entities for vehicles:
 - Buttons - horn, flash lights, wake up<sup>1</sup>, force data update<sup>1</sup> and trigger HomeLink. **Note:** The HomeLink button is disabled by default as some vehicles don't have this option. Enable via configuration/entities if desired.
 - Climate - turn HVAC on/off, set target temperature, set preset modes (defrost, keep on, dog mode and camp mode).
 - Device tracker - car location<sup>1</sup>.
-- Locks - door lock, rear trunk lock, front trunk (frunk) lock and charger door lock.
+- Locks - door lock, and charge port latch lock.
 **Note:** Set `state` to `heat_cool` or `off` to enable/disable your Tesla's climate system via a scene.
 - Selects - seat heaters and cabin overheat protection<sup>2</sup>.
 - Sensors - battery level, charge rate, energy added, inside/outside temperature, odometer and estimated range.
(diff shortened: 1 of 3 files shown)


alandtse:
Any idea why it thinks you changed the entire readme? I'd prefer the commit history was accurate.

InTheDaylight14:
I'm not sure why that happened. I can submit another PR

alandtse:
You don't have to close it. Just do a new commit that fixes the changes to the readme so it's only your changes.

InTheDaylight14:
> You don't have to close it. Just do a new commit that fixes the changes to the readme so it's only your changes.

I'll be honest, I don't know how to do that. After some research I tired the ```git restore``` on a version of the readme from the main repository dev branch. It looks like, after committing the restored readme, it is still -1,112 +1,112. Is there a different git command I should be using? I can do more research

alandtse:
@shred86 are you saying we should revert this?

shred86:
Doesn’t really matter to me. It’s calling the same endpoint as the charger port cover entity. I described what’s happening in [this post](https://github.com/alandtse/tesla/issues/285#issuecomment-1287812635) because there is one small “issue” where you can’t use the entity itself to call the cover open if it’s already open, so you have to use the cover.open_cover service. If folks are needing this for an automation, then it’s simple enough to call the cover.open_cover service.

InTheDaylight14:
The cover could be considered closed when the physical door is closed OR when the latch is engaged (The latch actualy says engaged when the door is closed anyway). That way the open door and unlatch functionality is available through the UI for both cases and through one entity. Not sure if that would be confusing to ppl if they are charging and the charge port door is "closed" in the UI.

alandtse:
I don't have a dog in this fight as I don't use this feature at all. If the new convention is to use a cover and people with an issue just have to do a manual service call, I don't really care. However, since @shred86 literally removed this functionality, I need to know whether you think this is something I should revert or not. Otherwise I'll have to deal with the same code getting removed and readded multiple times by different people.

Please decide between the two of you what we're doing here. If it's really separate functionality, we can keep the lock separate.

PR #338, opened 2022-11-09
feat: add window cover, attempt 2

InTheDaylight14:
Okay, I removed the split at ```is_closed```. I don't think it hurts to call ```close_windows``` if the windows are already closed and vice versa. With that split removed I'm expecting the function call to be asserted every time and the test to pass, independent of current window state. 

Also, I will update the readme through the online editor.

I apologize for the spam around this PR. I think I'm getting the hang of contributing. 

https://github.com/alandtse/tesla/issues/234


alandtse:
Normally you shouldn't keep closing and opening new PRs. Push your changes to the PR branch you're working on.

Issue #326, opened 2022-11-03
Deprecated function warning - is_metric


drlauridsen:
You might already be aware of this, but latest HA update shows this warning... just to make sure we are ready for future versions..

Logger: homeassistant.helpers.frame
Source: helpers/frame.py:77
First occurred: 09.31.43 (1 occurrences)
Last logged: 09.31.43

Detected integration that accesses the `is_metric` property of the unit system. This is deprecated and will stop working in Home Assistant 2023.1. Please adjust to use instance check instead.. Please report issue to the custom integration author for tesla_custom using this method at custom_components/tesla_custom/base.py, line 71: if self.hass.config.units.is_metric

SylvainGa:
Hi, as this fix been moved to the main branch? I installed that integration last weekend and noticed that warning today.

alandtse:
No. This will be moved when it actually may impact people. No sense forcing people to upgrade HA because of a deprecation warning.

SylvainGa:
Make sense. Not much time left though. According to the message in the log, it will stop working in Home Assistant 2023.1

tubalainen:
Looks like a dupe: https://github.com/alandtse/tesla/issues/442

PR #343, opened 2022-11-11
fix: #326 Replace deprecated code

zestysoft:
Tested in HA 2022.11.2 container -- no errors in logs, and I can see things like temp correctly for the vehicle.


alandtse:
Thanks. Can you please check when this new terminology was introduced in HA and whether we need to bump our minimum supported version?
https://github.com/alandtse/tesla/blob/dev/hacs.json#L4

zestysoft:
Looks like the change was just made in 2022.11.2:  https://github.com/home-assistant/core/commit/67d1dde69fbacf33f2c39ea14d89f2afa425ed18

alandtse:
Actually, it was 2022.11.0. You can expand the list to see and it looks like it came in via the beta.
image

alandtse:
Don't forget to `poetry update` after changing pyproject. We need the lock file committed too.

zestysoft:
Can `poetry update` and the lock file updates be a github action?

alandtse:
> Can `poetry update` and the lock file updates be a github action?

Maybe. But that updates all dependencies so not sure I want that done without some human testing.

zestysoft:
I was thinking the action would only run during the PR process?  Like before the merge and before the automated tests. A human could still pull the rip cord before the merge?

alandtse:
Happy to consider a PR for the automation. Regardless, this specific PR will probably hold until more people adopt 2022.11 and we actually get closer to the deprecation date.

alandtse:
As we're getting closer to the cutoff, I'll plan to merge this in soon.

alandtse:
Ok, just pushed a new release. Next release will include this. I'll plan for probably after xmas.

PR #351, opened 2022-11-17
fix: charge power sensor from int to float

emaberga81:
I would like the charger power sensor to be a float instead of an integer. I use it to calculate charging costs and the approximation to an integer makes the calculation less precise.

### DIFF
--- a/custom_components/tesla_custom/sensor.py
+++ b/custom_components/tesla_custom/sensor.py
@@ -175,7 +175,7 @@ class TeslaCarChargerPower(TeslaCarEntity, SensorEntity):
         self._attr_native_unit_of_measurement = POWER_KILO_WATT
 
     @property
-    def native_value(self) -> int:
+    def native_value(self) -> float:
         """Return the charger power."""
         return self._car.charger_power
 


custom_components/tesla_custom/sensor.py
```diff
@@ -175,7 +175,7 @@ def __init__(
         self._attr_native_unit_of_measurement = POWER_KILO_WATT
 
     @property
-    def native_value(self) -> int:
+    def native_value(self) -> float:
```
alandtse, line 178: This a typing hint and actually won't change anything. Did you test to confirm it actually does anything?

alandtse (changes requested):
Thank you for the PR. Please note the comment. Also, we try to add tests when possible so future PRs don't break anything. In this case, the test should be checking that a float is returned from this sensor based on the mock data. I don't know if a test has been written you can copy, but examples are here. https://github.com/alandtse/tesla/blob/dev/tests/test_sensor.py

emaberga81:
Hi, I did some test on my local installation and I changed the teslajsonpy car.py config. But I still get an approximated integer as value. I checked in the dwbug logs and to me it look like that is the value received by the car. In that case there is nothing to do. But I am not that advanced, I maight have missed something. Any tips?

alandtse:
I'd suggest enabling debugging on teslajsonpy and looking at what the API actually gives us. If it's an integer, we may be stuck since that's a Tesla issue. The [API docs](https://tesla-api.timdorr.com/) may give enough info to guess, although that may be out of date.

purcell-lab:
> I would like the charger power sensor to be a float instead of an integer. I use it to calculate charging costs and the approximation to an integer makes the calculation less precise.

Unfortunately the Tesla API only reports this value as an integer, which isn't very precise.

Maybe worth including as a standard sensor, given the focus on energy and power in HA.

I use a template to calculate Power = amps * voltage * phases

```yaml
    ev_power:
      friendly_name: "EV Power"
      value_template: "{{ (states('sensor.ev_phases')|int(0)
                          *state_attr('sensor.duka_charger_power','charger_volts')|int(0)
                          *state_attr('sensor.duka_charger_power','charger_actual_current')|int(0))}}"
      unit_of_measurement: W
      device_class: power

    ev_voltage:
      friendly_name: "EV Voltage"
      value_template: "{{ state_attr('sensor.duka_charger_power','charger_volts')}}"
      unit_of_measurement: V
      
    ev_current:
      friendly_name: "EV Actual Current"
      value_template: "{{ state_attr('sensor.duka_charger_power','charger_actual_current')}}"
      unit_of_measurement: A
      
    ev_charge_current_request:
      friendly_name: "EV Current Request"
      value_template: "{{ state_attr('sensor.duka_charger_power','charge_current_request')}}"
      unit_of_measurement: A
      
    ev_charging_time_left:
      friendly_name: "EV Charging Time Left"
      value_template: "{{ state_attr('sensor.duka_charging_rate','time_left')|int(3)}}"
      unit_of_measurement: hr
      
    ev_phases:
      friendly_name: "EV Charger Phases (adjusted)"
      value_template: "{{ max(1,(state_attr('sensor.duka_charger_power','charger_phases')|int(0)*3/2)|int(0))}}"
```

alandtse:
Since the API doesn't allow for this. I'll close this.

Issue #359, opened 2022-11-19
Vehicle battery sensor doesn't match car/app, uses SOC instead of uSOE


carleeno:
## Version of the custom_component

3.1.0

## Describe the bug

Noticeable in the colder months: the battery level through this integration does not match the app or car's reported battery level. This is because this integration appears to be using the battery SOC (state of charge) rather than the uSOE (usable state of energy).
The SOC doesn't change in cold weather as the amount of charge in the battery is technically the same, but the usable energy decreases in low temperatures. This is why the battery values shown in the app and in the car display use the uSOE value which reflects that less energy is available (along with the snowflake icon).

The Vehicle State API has uSOE available as `charge_state.usable_battery_level`, I'm happy to create a PR (in teslajsonpy as well) but wanted to start the discussion if the main battery sensor should match the car/app, or if a second sensor for usable battery should be created instead. We could even add a snowflake attribute that sets true when there is >= 2% difference between the 2 values to also match the car/app behavior.

alandtse:
Thanks for bringing this up. I don't live in a cold climate so wouldn't have known about this. Given that I don't think HA has a native way to address this, I think it's really our choice on what makes the most sense for users. I'd lean toward matching the app behavior to avoid confusion. For advanced users, we could expose an additional sensor or attribute. That could also be gated by an option flag.

If you want to just propose something that makes sense to you, we can leave it in an open PR for a little bit before integrating to see what the reaction is. But to be honest, it may be best to just release and see if anyone really complains one way or the other before changing it.

carleeno:
awesome, thank you, I'll spin a pr to update the existing battery sensor to reflect the app/car (uSOE); and add a "raw_soc" attribute for those who may want to show if their battery is cold for example (I personally would then make a template sensor to add a snowflake icon to my dashboard)

alandtse:
I'll hold off on the next release to get this in too.

carleeno:
thanks, will create the pr shortly

PR #362, opened 2022-11-20
fix: vehicles use usable_battery_level to match app

carleeno:
For #359 

This updates the battery sensor for vehicles to use the `usable_battery_level` value so that it matches the Tesla app and in-car display in all weather conditions.

Additionally a `raw_soc` attribute is added to the battery sensor to provide insight into the battery's SOC, which does not change with temperature. This allows for users to detect reduced range due to cold (such as the snowflake icon in the app) if they wish to do so.

Lastly this PR fixes a bug in that the battery icon did not reflect Charging status due to a typo that I happened to notice while doing the above.

### DIFF
--- a/custom_components/tesla_custom/manifest.json
+++ b/custom_components/tesla_custom/manifest.json
@@ -5,7 +5,7 @@
   "documentation": "https://github.com/alandtse/tesla/wiki",
   "issue_tracker": "https://github.com/alandtse/tesla/issues",
   "requirements": [
-    "teslajsonpy==3.1.0"
+    "teslajsonpy==3.2.0"
   ],
   "codeowners": [
     "@alandtse"
--- a/custom_components/tesla_custom/sensor.py
+++ b/custom_components/tesla_custom/sensor.py
@@ -99,16 +99,24 @@ class TeslaCarBattery(TeslaCarEntity, SensorEntity):
     @property
     def native_value(self) -> int:
         """Return battery level."""
-        return self._car.battery_level
+        # usable_battery_level matches the Tesla app and car display
+        return self._car.usable_battery_level
 
     @property
     def icon(self):
         """Return icon for the battery."""
-        charging = self._car.battery_level == "Charging"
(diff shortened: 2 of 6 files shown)


carleeno:
@alandtse this test failure is weird, the same tests pass on my local branch, and none of the mock data or tests have changed for `charge_rate` in either this repo or teslajsonpy afaict. What makes it weirder is that 37.3 is almost the km equivalent of 23.2 miles, but it's off by more than a rounding error, so it doesn't seem like it's getting converted somewhere.

```
    async def test_charger_rate_value(hass: HomeAssistant) -> None:
        """Tests charger_rate is getting the correct value."""
    
        state = hass.states.get("sensor.my_model_s_charging_rate")
>       assert state.state == str(car_mock_data.VEHICLE_DATA["charge_state"]["charge_rate"])
E       AssertionError: assert '37.3' == '23.2'
E         - 23.2
E         + 37.3

tests/test_sensor.py:195: AssertionError
```

Should I just push a noop commit to re-kick the tests and see if it was somehow a fluke? Or do you have an idea of what's causing this?

carleeno:
I saw that I forgot to rebase, maybe that was somehow why...let's see if it passes this time...

alandtse:
I mean that's an odd failure but I really want to avoid pushing commits that cause the tests to fail randomly like that. Is it trying to convert the values?

carleeno:
I'm continuing to dig into this today. Yes for some reason the integration is converting the charge rate from 23.2 mph to 37.3 km/h which is breaking the test. I'm not able to reproduce at all testing locally so I'm using github actions in my own forked repo to try to reproduce and narrow it down.

carleeno:
ok I was able to reproduce this with simply a poetry update from dev and no other changes.

It appears some existing tests fail with the newest version of home assistant. Since it's not related to this change, I'll create a new PR to fix the tests.
For this PR I reverted the poetry update and only updated teslajsonpy.

PR #363, opened 2022-11-20
fix: Update tests for hass built-in unit conversions in 2022.11

carleeno:
Updating homeassistant brought about failures in existing tests due to hass automatically converting units in sensors from miles to km (and speed equivalents).

There doesn't seem to be a way to globally disable unit conversion for tests, so this PR fixes failing tests by converting the mock values if necessary. This feels like a hack, so I'm open to suggestions. I alternatively tried setting `hass.config.units` to imperial, but that breaks the temperature tests since it converts the mock Celsius temps into Fahrenheit, so a conversion of the mock values would be needed regardless.


alandtse:
I wonder if we need to update the ha custom component helper....

PR #365, opened 2022-11-21
fix: don't remove entities when reloaded while car is asleep

### LINKED ISSUE #293: Upgrade 3.0 sentry mode switch missing
After installing the new 3.0 integration, it seems the sentry mode switch is missing?

carleeno:
This ensures the following entities aren't suddenly removed if the integration is reloaded (or HA restarted) while a vehicle is asleep:
- heated steering wheel
- sentry mode
- homelink

Instead of conditionally creating these entities during setup, they will always be added. However, on first install (when wake_if_asleep is used), if any of these entities are not available they will be disabled by default. This ensures that entities which are not part of the vehicle will not be enabled (not visible and no logging), but also provides the option that a user could enable it if it was somehow wrongly disabled.

Once enabled, the entity would not self-disable upon reload even if it is unavailable: Fixes #293 (note these entities would show unavailable if reloaded while the vehicle is asleep, but would automatically become available again next time it's online)

Additionally this PR fixes dynamic updates of the availability of these items while the car is online (and reflect the last state while car is offline).


carleeno:
@shred86 for vis, this should fix the remaining entities disappearing issues, this seemed the next best thing given the option codes aren't reliable.

alandtse:
Back out the `is_metric` change please as there is an existing pr and I don't want to force a ha min update yet since that change is very recent.

carleeno:
> Back out the `is_metric` change please as there is an existing pr #343 and I don't want to force a ha min update yet since that change is very recent.

Ahh I didn't realize it was so recent. Done.

alandtse:
Since this is an area with active issues, can you also put in some tests covering these sensors being created with tests for both cases that cause then to be enabled or disabled?

carleeno:
> Since this is an area with active issues, can you also put in some tests covering these sensors being created with tests for both cases that cause them to be enabled or disabled?

Took me a bit to work this out, let me know how that looks.

alandtse:
Looks good. Thanks for that.

Issue #356, opened 2022-11-18
Car Software Update is showing Installing when Scheduled


alexbeatle:
![image](https://user-images.githubusercontent.com/56932253/202814076-68597346-5f87-4765-8f1d-9e107ac13254.png)

The Car Software update entity is showing as "installing" when it's actually been scheduled to be installed later in the day.

alandtse:
We'll need the `teslajsonpy` output for the update info to correct this. Can you please enable debugging on it and see if we can capture the value when scheduled?

carleeno:
In `update.py` `in_progress()` it specifically returns `True` when `update_status == "scheduled"`, it looks like it was intentional?

Last night/this morning I have an update available, so I've been collecting response data and tweaking the update component to show the different states (Waiting on Wi-Fi, Downloading, Available, Scheduled, etc) and only enable the Install button when it's able to be installed (Available and Scheduled).

Since it's possible to install immediately while scheduled for a later time, I think it's better to not show it as in-progress while scheduled so that the install button is still available. I'm thinking it could look something like this?
![image](https://user-images.githubusercontent.com/54586926/203565029-6e1e5f23-611e-410d-81a8-cdc8d6bb66f0.png)

KirkKirk:
This is a great option. I'm getting notifications even before the Tesla app.
How can I get info when the software is downloaded and ready to install?
I need it to create automation to trigger the installation in specified conditions (i.e. if received at night and there is enough time before going to work).

carleeno:
> This is a great option. I'm getting notifications even before the Tesla app. How can I get info when the software is downloaded and ready to install? I need it to create automation to trigger the installation in specified conditions (i.e. if received at night and there is enough time before going to work).

This example should work because it will try to install whenever the status string changes in the latest version attribute...but it will only succeed to start the installation when it's ready to install since the install feature is only enabled when it's finished downloading or scheduled. Obviously you'd want to probably add more conditions, such as at home, parked, enough charge, etc.

```yaml
description: "Auto Update"
mode: single
trigger:
  - platform: state
    entity_id:
      - update.mystique_software_update
    attribute: latest_version
condition:
  - condition: time
    before: "05:00:00"
    weekday:
      - mon
      - tue
      - wed
      - thu
      - fri
action:
  - service: update.install
    data: {}
    target:
      entity_id: update.mystique_software_update
```

[...]

PR #370, opened 2022-11-23
feat: show software update statuses in version info

carleeno:
I had a software update last night, so I seized the opportunity to collect api responses of the different states, and provide some improvements to the update component.

This makes a few improvements:
- Ensures the install button isn't available until the update is downloaded (it is hidden when waiting on wifi or still downloading)
- Ensures the install button is available when the update is scheduled, in case the user wants to install immediately.
- Shows the status (Downloading, Waiting on Wi-Fi, etc) in the latest version string to indicate to the user why the install button is not yet available, or to show if it's already scheduled.
- No longer shows as "Installing" when it's scheduled, #356 

Some examples
image

image

image

image

image

### DIFF
--- a/custom_components/tesla_custom/update.py
+++ b/custom_components/tesla_custom/update.py
@@ -27,6 +27,17 @@ async def async_setup_entry(hass: HomeAssistant, config_entry, async_add_entitie
     async_add_entities(entities, True)
 
 
+INSTALLABLE_STATUSES = ["available", "scheduled"]
+
+PRETTY_STATUS_STRINGS = {
+    "downloading_wifi_wait": "Waiting on Wi-Fi",
+    "downloading": "Downloading",
+    "available": "Available to install",
+    "scheduled": "Scheduled for install",
+    "installing": "Installing",
+}
+
+
 class TeslaCarUpdate(TeslaCarEntity, UpdateEntity):
     """Representation of a Tesla car update."""
 
@@ -43,7 +54,14 @@ class TeslaCarUpdate(TeslaCarEntity, UpdateEntity):
     @property
     def supported_features(self):
         """Return the list of supported features."""
-        return UpdateEntityFeature.INSTALL | UpdateEntityFeature.PROGRESS
+        # Don't enable Install button until the update has downloaded
+        if (
+            self._car.software_update
+            and self._car.software_update.get("status") in INSTALLABLE_STATUSES
+        ):
+            return UpdateEntityFeature.INSTALL | UpdateEntityFeature.PROGRESS
+        else:
+            return UpdateEntityFeature.PROGRESS
 
     @property
     def release_url(self) -> str:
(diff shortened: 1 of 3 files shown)


tests/test_update.py
```diff
@@ -27,28 +27,27 @@ async def test_software_update_properties(hass: HomeAssistant) -> None:
     version = car_mock_data.VEHICLE_DATA["vehicle_state"]["software_update"]["version"]
 
     state = hass.states.get("update.my_model_s_software_update")
-    # assert state.state == str(car_mock_data.VEHICLE_STATE["software_update"])
-
-    assert state.attributes.get("latest_version") == version
+    assert state.state == "on"
+    assert state.attributes.get("latest_version") == version + " (Available to install)"
```
alandtse, line 31: Do you mind adding a test confirming #356 is fixed? I know your fix addressed it just by looking at the code, but I like to get tests in if they were previously reported issues so we don't get a regression in the future.

carleeno: Can do, I was hoping to iterate through each status in a for loop to test everything, but I'm failing to get the state to update mid-test after updating the mock data. Not sure if something to do with the coordinator or what.

something like this, but even re-running setup_platform doesn't update the values after the first time

```python
    for raw, pretty in status_strings.items():
        car_mock_data.VEHICLE_DATA["vehicle_state"]["software_update"]["status"] = raw
        # Something here to update states?
        state = hass.states.get("update.my_model_s_software_update")
        assert state.attributes.get("latest_version") == f"{version} ({pretty})"
        # assert in-progress, installable, etc...
```

If you have any tips I'm all ears. I'll keep digging though.

alandtse: Ok the issue is you're just updating the mock data which is basically the API. You haven't ingested it into the component yet unless you call an update on the component to reach out to the `API`. Other tests are doing it by just setting up the component again which forces a `pull` from the API using the new mocked data.

You could try looking at HA tests to see how people are doing updates. That said, I can live with the idea of you changing the mock data, and then having the entity setup and then test the value. That at least covers all the interesting startup states.

carleeno: > Ok the issue is you're just updating the mock data which is basically the API. You haven't ingested it into the component yet unless you call an update on the component to reach out to the `API`. Other tests are doing it by just setting up the component again which forces a `pull` from the API using the new mocked data.
> 
> You could try looking at HA tests to see how people are doing updates. That said, I can live with the idea of you changing the mock data, and then having the entity setup and then test the value. That at least covers all the interesting startup states.

After looking at several example tests in the HA repo, it seems what I was trying to do isn't very typical. So I went the route to setup for each state.
Unfortunately doing a setup twice in the same test function didn't overwrite the previous state either, so I created a test function for each phase of the update status. It's probably a little overkill now tbh.

alandtse: Ok thanks.

custom_components/tesla_custom/update.py
```diff
@@ -27,6 +27,17 @@ async def async_setup_entry(hass: HomeAssistant, config_entry, async_add_entitie
     async_add_entities(entities, True)
 
 
+INSTALLABLE_STATUSES = ["available", "scheduled"]
+
+PRETTY_STATUS_STRINGS = {
```
alandtse, line 32: I want to think about this more on whether we should hard code language strings in. I'm ok with this for now.

Issue #160, opened 2022-03-15
Tire Pressure Sensors


giggo:
Apparently the tire pressure is now available via the Tesla API ( [https://www.notateslaapp.com/de/software-updates/version/2022.4.5/versionshinweise](https://www.notateslaapp.com/de/software-updates/version/2022.4.5/versionshinweise) ). Are there plans to add them as sensors to HA?

alandtse:
People are welcome to submit PRs to add features.

k-korn:
TPMS data is available in the following attributes of "vehicle_state" :
tpms_pressure_fl
tpms_pressure_fr
tpms_pressure_rl
tpms_pressure_rr

However, quick look on https://github.com/alandtse/tesla/blob/dev/custom_components/tesla_custom/sensor.py suggests that [teslajsonpy ](https://github.com/zabuldon/teslajsonpy) should also be modified to add this data.

k-korn:
With version 2.0 it is now possible to use now exposed vehicle_data to display TPMS sensors in HASS.
- on Config/Devices page, enable "sensor.model_3_vehicle_state_data_sensor" in Tesla integration, which is disabled by default
- Create 4 Template sensors as follows for FL, FR, RL, RR sensors:
```
 template:
    sensors:
     - name: "Tesla TPMS FL"
       state: >
        {% set tpms = state_attr("sensor.model_3_vehicle_state_data_sensor", "tpms_pressure_fl") %}
        {{ tpms|round(2) if tpms != None else states('sensor.tesla_tpms_fl') }}
       unit_of_measurement: "PSI"
```
Extra steps with intermediate variable were added to remove "Unavailable" TPMS sensor state values and keep actual values visible.
Then we can use something like a Grid Lovelace card to display tire pressures: https://i.imgur.com/4gpfwJj.png

alandtse:
Thanks for providing the template info. Do you mind creating a wiki entry on it too so when I close this thread people can find it?

k-korn:
Sure: https://github.com/alandtse/tesla/wiki/TPMS-Data-Display

k-korn:
Version 3.0 of the integration removed vehicle_state_data_sensor data, so the solution above can no longer be applied.

Issue #294, opened 2022-10-22
v3.0 missing old state sensors


FalconUK:
I used to extract tyre pressures from the attributes of 

'vehicle_state_data_sensor'

This sensor appears to be missing.

There was also a similar sensor 'drive_state_data_sensor', or something like that?

Are these sensors coming back?

alandtse:
@shred86 I didn't realize we took them away. Any thoughts on whether we bring them back? These were the raw json sensors for people to template on

shred86:
I'm not too familiar with the previous version as I honestly only started looking into this integration with the rewrite version. Was it just a sensor with a bunch of attributes from the Tesla API vehicle_data endpoint? Seems like it would be a very long list of attributes... My thought is we should just add the entities people are using so they don't have to create template sensors. Tire pressures _should_ be an easy one to create sensors for, but we would have to add it on the teslajsonpy side as I don't think I made those into properties.

FalconUK:
@shred86, @alandtse, I was using tyre pressures to nurse a slow puncture through until a replacement tyre arrived. I declared all four, example below:

```
    - sensor:
        - name: "*** tyre pressure front left"
          unique_id: sensor.***_tyre_pressure_front_left
          state_class: measurement
          state: "{{(state_attr('sensor.***_vehicle_state_data_sensor', 'tpms_pressure_fl')|float * 14.5038)|round(2)}}"
          unit_of_measurement: "PSI"
```

This raises a similar localisation question as with the units for the odometer sensor - km/miles, bar/PSI

It's fair to say that some attributes are definitely more useful than others. I think it would be in line with the current philosophy to expose the majority of sensors as entities, perhaps with some of them disabled by default?

alandtse:
Yes, the intent of the rewrite was to make it much easier to edit the ha component for contributors. If it's an entity that is important, it should be created. Ideally we can templatize it so it's just data driven.

craigsbits:
The state sensors were useful the other day when the battery state of charge disappeared!  Personally I would like to see them back especially the tyre (tire) pressures back.  Support the odometer units mention above.  The heated seats in the official app now have an "auto" setting if that could be found and added.  I will open a new thread for that.

bkr1969:
So is the "old" version completely out the window?  It seems poor practice to rewrite an integration that was working just fine to one that doesn't work at all.

TheDK:
[...]

PR #376, opened 2022-11-28
feat: add tpms pressure sensors

InTheDaylight14:
Mentioned in https://github.com/alandtse/tesla/issues/294

closes https://github.com/alandtse/tesla/issues/160

PR adds pressure sensors for the four TPMS tire sensors.


custom_components/tesla_custom/sensor.py
```diff
@@ -497,3 +507,50 @@ def native_value(self) -> Optional[datetime]:
...
+        self._attr_device_class = SensorDeviceClass.PRESSURE
+        self._attr_state_class = SensorStateClass.MEASUREMENT
+        self._attr_native_unit_of_measurement = PRESSURE_BAR
+        self._attr_icon = "mdi:gauge-full"
+        # self._value: Optional[datetime] = None
+
+    @property
+    def native_value(self) -> float:
+        """Return TPMS Pressure."""
+        if self._tpms_sensor == "TPMS front left":
```
alandtse, line 534: This use of a string compare is brittle particularly when it's defined above as a constant. Why not replace the list with a dict with the key being the text string and the value being the attribute we want found?

That way:
```python
    return getattr(self._car, "TPMS_SENSORS.get(self._tpms_sensor))
```

alandtse: We can also add a default case where if there is no attribute, we return None or something.

InTheDaylight14: Very cool, I was looking for a way to do that! Thanks

custom_components/tesla_custom/sensor.py
```diff
@@ -19,6 +19,7 @@
     SPEED_MILES_PER_HOUR,
     TEMP_CELSIUS,
     TIME_HOURS,
+    PRESSURE_BAR,
```
alandtse, line 22: Black should be alphabetizing this. Are you sure you're running it?

FalconUK:
Great to see these added. A small point, and it's a picky one, but I think the friendly name should have TPMS all in upper case, rather than just the T.

Issue #310, opened 2022-10-26
Door States


GitHubGoody:
If the doors' states are available in the API like the Tesla app, it would be nice to have those as entities as well.

shred86:
It looks like they are. I'm trying to figure out the best HA entity to use for implementation. Any Model X owners able to confirm if you are or are not able to open your car door using the Tesla app? If there is the capability to open the door via then app, then we need to figure out the API endpoint to use.

PR #377, opened 2022-11-28
feat: add doors binary sensor

InTheDaylight14:
closes https://github.com/alandtse/tesla/issues/310


custom_components/tesla_custom/binary_sensor.py
```diff
@@ -205,3 +206,47 @@ def __init__(
...
+        """Initialize car door entity."""
+        super().__init__(hass, car, coordinator)
+        self.type = "doors"
+        self._attr_device_class = BinarySensorDeviceClass.DOOR
+        self._attr_icon = "mdi:car-door"
+
+    @property
+    def is_on(self):
+        """Return True if a car door is open."""
+        if (
```
alandtse, line 229: This should just be a return of the test.
```python
   return self._car.door_df or ... 
```

Issue #388, opened 2022-12-01
Tire pressure in PSI


smarthomeaf:
**Is your feature request related to a problem? Please describe.**
Would we be able to also get the Tire Pressure in PSI in addition to bar.  In the US that is a well know representation and we typically don't understand bar.   Maybe ths could just be an attribute.   I know I can create a manual template sensor by doing some calculations, but this would no apply to most of the HASS population
**Describe the solution you'd like**

An attribute to the existing Sensor should be enough, but would prefer to choose have the option int eh config on how I woudl like to represent tire pressure

**Describe alternatives you've considered**
Creating a template sensor to calculate PSI via a Calculation

**Additional context**

carleeno:
Hi, fyi you can change the units of any sensor, and psi is available. Click on the TPMS sensor, click settings, and you will see the option
![image](https://user-images.githubusercontent.com/54586926/205120673-ed607a4f-1fa3-4b04-b809-1f7a48404ff4.png)

There is probably still some discussion of if we should make the default to be psi instead of bar though. I would think that HA should specify the default based on locale, but it seems not to do so for pressure.

alandtse:
If it's possible to do a default, we should do it based on the app. However, the question is whether that is possible in HA at all.

InTheDaylight14:
Yeah using ```native_unit_of_measurement``` enables the user to select their preferred unit of pressure via the UI. I just tried to code ```unit_of_measurement``` or ```suggested_unit_of_measurement``` to PSI alongside the ```native_unit_of_measurement``` but that did not work for defaulting the value to PSI over Bar in the front end. 

The value comes to us as Bar but we can always do the conversion to PSI, round to 1 decimal point, and have PSI be the ```native_unit_of_measurement```.

Interesting that the US app shows the values in PSI and the API has the values in Bar.

alandtse:
> I just tried to code unit_of_measurement or suggested_unit_of_measurement to PSI alongside the native_unit_of_measurement but that did not work for defaulting the value to PSI over Bar in the front end

Wait. That sounds like a HA bug. Can you please do a quick check if it should be reported?

PR #391, opened 2022-12-01
fix: default pressure from Bar to PSI

InTheDaylight14:
closes https://github.com/alandtse/tesla/issues/388


InTheDaylight14:
The rounding of the native to suggested unit conversion does not make any sense to me. I don't want to write a flimsy test not understanding how that works. 

No mention of significant digits or rounding in the [unit_conversion utility](https://github.com/home-assistant/core/blob/dev/homeassistant/util/unit_conversion.py#L65-L90).

carleeno:
> The rounding of the native to suggested unit conversion does not make any sense to me. I don't want to write a flimsy test not understanding how that works.
> 
> No mention of significant digits or rounding in the [unit_conversion utility](https://github.com/home-assistant/core/blob/dev/homeassistant/util/unit_conversion.py#L65-L90).

I believe the precision scaling happens [here](https://github.com/home-assistant/core/blob/dev/homeassistant/components/sensor/__init__.py#L785-L795), I only found this because I was digging for where the unit conversion happens some time ago

InTheDaylight14:
Oh nice find, thanks!

Issue #413, opened 2022-12-08
sensor.vehicle_energy_added wrongly counting when resuming charging


evolucja:
![image](https://user-images.githubusercontent.com/990929/206454345-ae0bc0c0-d65c-4758-a66d-70fdb48615cd.png)

Hi, it looks like when we stop charging and then start it again (or when it reached certain percentage and then we increase the percentage so it starts charging again without starting the new session) the counter doesn't start from 0 but from the last value so using it for energy meters can double or triple the charged kWh amount.

Any idea how to fix that?

evolucja:
Unfortunately, it still doesn't work with 3.7.0. The car was charged last a few days ago. Since then it counted energy added 5 times.

![image](https://user-images.githubusercontent.com/990929/208134205-bfd2dc7f-c42c-4af1-9fe8-5efb84911738.png)

carleeno:
@evolucja can you share the same history time range of the
- charging sensor
- charger sensor
- online sensor
- alseep sensor

we need to correlate when it is going to 0 to try to figure out why it's doing that. I double checked the code and we're not manually setting 0 for that sensor anywhere in the code so now I'm wondering if the car is reporting 0 sometimes.

evolucja:
@carleeno here it is:

![image](https://user-images.githubusercontent.com/990929/208141709-11c72e43-fe79-4578-a560-2eeca43a47a3.png)

carleeno:
@evolucja almost, it's supposed to reset the energy added value when charger goes from 0 to 1, and retain that value when the charger goes from 1 to 0. Here's mine so you can see what it's supposed to look like
![image](https://user-images.githubusercontent.com/54586926/208221727-35f3411d-8113-4b18-a24b-a656c72a215d.png)

it looks like when you unplug the charger and plug it back in, it's resetting to 0, but then you never charge and unplug it again, and since it never charged it reset to the previous energy added value. I'll try to replicate this with my car by stopping the charge, unplugging, plugging back in, and unplugging before it charges again.

Note the data you are seeing is directly coming from Tesla's api, we are no longer manipulating the value. If there is a value available, we give it; if there is no data available, it shows as `unknown`...so I'm not sure there is anything we can do about this behavior if tesla is resetting the value to 0 and then back to the previous value. Maybe you can also help verify this with debug logs and try to reproduce using the same steps, recording the data json each time.

evolucja:
Hmm don't know what happened but it normalized after the newest update and so far looks good...

[...]

PR #414, opened 2022-12-08
fix: always show charge energy for current and last session

### LINKED ISSUE #295: V3 sensor.vehicle_energy_added incompatible with energy dashboard

3.0.0

## Describe the bug

I understand the vehicle _energy_added sensor is a bit of a cludge. I'm raising this issue to highlight it's incompatibility with the HA energy dashboard.

Here is my sensor for today:

Screenshot_20221023-111728

Looks like I have consumed about 35 kWh for the day.

Energy dashboard is however reporting 135 kWh for the day.

Screenshot_20221023-111841

It is doing this as the energy sensor is resetting to 0 between charges and the energy dashboard double counts every time it rests to zero.

I suspect it maybe required to make a smarter energy sensor to support this integration and the HA energy dashboard.

carleeno:
Fixes #295 and #413 

Based on the information [here](https://developers.home-assistant.io/docs/core/entity/sensor/#entities-representing-a-total-amount), I realized that the logic to set the energy added to 0 when not charging was breaking energy calculations.

The actual energy added value is reset by tesla on it's own, so when we manually set 0, and then revert back to the tesla's value if charging starts again for the same charge session, it will cause the energy dashboard to incorrectly sum the deltas.

It appears that TOTAL_INCREASING is the right state class to use, as we don't expect the value to decrease substantially (a 10% allowance is built in). So if we just rely on the car's own charging state to reset appropriately, the energy dashboard should reflect usage correctly.

What this changes for users:
- The energy added sensor will continue to report the last energy added value from the most recent charging session
- The energy dashboard will no longer double/triple the energy usage for single charging sessions which pause/resume
- The energy added sensor will show unknown instead of 0 when the data is not available (if the vehicle is offline/asleep when the integration is loaded)

### Test plan

Tested in dev container with a model 3:
- ensured while unplugged, the energy_added sensor still showed the last value from the previous charging session
- ensured that upon plugging in a charger, the energy_added sensor reset to 0
- ensured the energy_added sensor increased as it charged
- ensured that when pausing charging (remaining plugged in) the energy_added sensor did not reset to 0
- let the car sleep while plugged in then reloaded integration to get an "unknown" state for energey_added sensor
- woke the car up and ensured the energy_added sensor reflected the last value
- at the end of the charging session, ensured that the energy dashboard amount reflected the actual sensor

energy_added sensor showing that it did not reset to zero when pausing (the blank area is "Unknown" from reloading while asleep)
image

The energy dashboard value matches the sensor (it did not double it, even after the "unknown" period)
image

### DIFF
--- a/custom_components/tesla_custom/sensor.py
+++ b/custom_components/tesla_custom/sensor.py
@@ -164,9 +164,9 @@ class TeslaCarChargerEnergy(TeslaCarEntity, SensorEntity):
     @property
     def native_value(self) -> float:
         """Return the charge energy added."""
-        if self._car.charging_state == "Charging":
-            return self._car.charge_energy_added
-        return "0"
+        # The car will reset this to 0 automatically when charger
+        # goes from disconnected to connected
+        return self._car.charge_energy_added
 
     @property
     def extra_state_attributes(self):


carleeno:
in draft while I test that the `unknown` behavior doesn't cause any problems in the energy dashboard

evolucja:
I was about to give a try implementing this change but many thanks you've done it!  :)

evolucja:
Unfortunately @carleeno, it still doesn't work with 3.7.0. The car was charged last a few days ago. Since then it counted energy added 5 times.

image

PR #419, opened 2022-12-11
fix: Charge time remaining, account for time passing between updates

InTheDaylight14:
Saves the time when the value changes, subtracts the time that has passed since the last update. 

Without this the value grows by 1 minute for every minute since an api update. Then, on api call, the time jumps back down to the proper time.

### DIFF
--- a/tests/test_sensor.py
+++ b/tests/test_sensor.py
@@ -261,6 +261,15 @@ async def test_time_charge_complete_charging(
     assert state.attributes.get(ATTR_DEVICE_CLASS) == SensorDeviceClass.TIMESTAMP
     assert state.attributes.get(ATTR_STATE_CLASS) is None
 
+    mock_timetravel = mock_now + timedelta(minutes=2)
+    monkeypatch.setattr(dt, "utcnow", lambda: mock_timetravel)
+
+    state = hass.states.get("sensor.my_model_s_time_charge_complete")
+    charge_complete = mock_now + timedelta(
+        hours=float(car_mock_data.VEHICLE_DATA["charge_state"]["time_to_full_charge"])
+    )
+    charge_complete_str = datetime.strftime(charge_complete, "%Y-%m-%dT%H:%M:%S+00:00")
+
 
 async def test_time_charge_completed(hass: HomeAssistant) -> None:
     """Tests time charge complete is the correct value."""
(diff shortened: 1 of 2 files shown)


alandtse:
For this type of nuanced issue, do you think you can write a test so we don't get a regression? Basically, the idea is the test will fail if this fix's behavior is undone.

InTheDaylight14:
Hmm, my first thought on how to test would be to put a delay of 1 minute in and make sure the state has not changed. I'd hate to hold up the all the tests for 1 minute though. 

I included this behavior in the Destination arrival time PR. So that could end up with 2 minute delays in the tests. 

Is there a way to programmatically change the system time? Or run the tests with delays in a separate thread?

alandtse:
Yes, use libraries like https://pypi.org/project/time-machine/. I don't remember what HA has standardized on though.

InTheDaylight14:
Moving to a draft for now. This might not be needed if the polling is increased to every 60 seconds while charging. The charge time completion will always be within 60 seconds. 

https://github.com/alandtse/tesla/issues/426

tests/test_sensor.py
```diff
@@ -261,6 +261,15 @@ async def test_time_charge_complete_charging(
     assert state.attributes.get(ATTR_DEVICE_CLASS) == SensorDeviceClass.TIMESTAMP
     assert state.attributes.get(ATTR_STATE_CLASS) is None
 
+    mock_timetravel = mock_now + timedelta(minutes=2)
```
InTheDaylight14, line 264: Little out of my element here, I want to draw attention to this to make sure this works for testing. Trying to change the time 2 minutes in the future and make sure that the state has stayed the same.

alandtse: I'm not certain what the issue is. Is the test failing? I haven't mocked timestamps before so I'd just be reading the documentation like you are and reviewing the HA examples.

InTheDaylight14: No the test is passing just fine. Maybe @carleeno can review the mock_timetravel. I based it off of their recent PR to shore up time tests. I just want to make sure that the state is actually being evaluated two minutes later by using that ```monkeypatch.setattr...```

carleeno: @InTheDaylight14 the mock looks correct, however you're not asserting anything afterwards so it's not actually testing anything.

Add an `assert state.state == charge_complete_str` at the end and see what happens, this will reveal if your logic is good.

InTheDaylight14:
Oh and it looks like the tests are being skipped in the checks. Not sure what is going on there.

tests/test_sensor.py
carleeno: for an explanation of what the monkeypatch is doing is it's replacing the `homeassistant.util.dt.utcnow` that HA uses with our mocked time. So when `dt.utcnow` is called from `sensor.py` it will actually get your mocked value.

PR #427, opened 2022-12-12
docs: update readme with new features

InTheDaylight14:
Adding a few new features to the list of entities.


alandtse:
Thanks. Please note that you may also want to start using semantic versioning for all commits. That avoids me accidentally merging the bad history in if I forget to squash. (which I did with your last PR). ;)

PR #440, opened 2022-12-18
fix: remove state class from timestamp sensors

carleeno:
I discovered while testing on HA dev branch that there is a new check to ensure only numeric sensors are reporting a state class (core PR 83344)

With this new check, it logs an error message that some of our sensors are non-numeric, but have a state class

```
[...]
  File "/usr/local/python/lib/python3.10/site-packages/homeassistant/components/sensor/__init__.py", line 797, in state
    raise ValueError(
ValueError: Sensor sensor.mystique_time_charge_complete has a state class and thus indicating it has a numeric value; however, it has the non-numeric device class: timestamp
```

This PR fixes this error, and also fixes the history graphs for these sensors to not try to map the timestamp to a line chart (which results in unreadable values on the chart as it's converting the timestamp to a float)

Additionally this PR fixes a flakey datetime test by mocking `dt.utcnow`

### DIFF
--- a/custom_components/tesla_custom/sensor.py
+++ b/custom_components/tesla_custom/sensor.py
@@ -501,7 +501,6 @@ class TeslaCarTimeChargeComplete(TeslaCarEntity, SensorEntity):
         super().__init__(hass, car, coordinator)
         self.type = "time charge complete"
         self._attr_device_class = SensorDeviceClass.TIMESTAMP
-        self._attr_state_class = SensorStateClass.MEASUREMENT
         self._attr_icon = "mdi:timer-plus"
         self._value: Optional[datetime] = None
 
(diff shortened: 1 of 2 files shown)


carleeno:
:( seems we have a flaky time-based test as well
I'll see if I can address in this PR too
```
=================================== FAILURES ===================================
__________________________ test_arrival_time[pyloop] ___________________________
[gw1] linux -- Python 3.10.9 /home/runner/work/tesla/tesla/.venv/bin/python

hass = <homeassistant.core.HomeAssistant object at 0x7f9c410586d0>

    async def test_arrival_time(hass: HomeAssistant) -> None:
        """Tests arrival time is getting the correct value."""
    
        state = hass.states.get("sensor.my_model_s_arrival_time")
        arrival_time = datetime.utcnow() + timedelta(
            minutes=round(
                float(
                    car_mock_data.VEHICLE_DATA["drive_state"][
                        "active_route_minutes_to_arrival"
                    ]
                ),
                2,
            )
        )
        arrival_time_str = datetime.strftime(arrival_time, "%Y-%m-%dT%H:%M:%S+00:00")
    
>       assert state.state == arrival_time_str
E       AssertionError: assert '20[22](https://github.com/alandtse/tesla/actions/runs/3725591685/jobs/6318493375#step:8:23)-12-18T16:[42](https://github.com/alandtse/tesla/actions/runs/3725591685/jobs/6318493375#step:8:43):54+00:00' == '2022-12-18T16:42:55+00:00'
E         - 2022-12-18T16:42:55+00:00
E         ?                   ^
E         + 2022-12-18T16:42:54+00:00
E         ?                   ^

tests/test_sensor.py:558: AssertionError
```

Issue #164, opened 2022-03-19
Scedule departue


Dominic-070:
a switch and time field for setting a scheduled time for charging and HEVC, as in the Tesla App

michaelsidenius:
That would be a good enhancement

InTheDaylight14:
@alandtse @shred86 Wanted to run my thoughts by y'all before building. I was originally thinking the following:
- Select for Off/Sch Charge/Sch Depart
- Select for Off/All days/Weekday Only preconditioning and off-peak 
- Input_datetime for Sch Charge/Sch Depart/Off-Peak End

However there are no input_datetimes available to custom integrations. [Discussion](https://community.home-assistant.io/t/time-date-input-entity/410973).

This has me thinking about a different path. Exposing a service call for setting scheduled charging and setting scheduled departure with user friendly argument descriptions.

I would also want to add various attributes to the climate entity for the state of scheduled charge/depart, the time, status of weekend only, etc.

alandtse:
Sounds good. Given there are other custom components who had to improvise, let's see if it makes sense to follow their service call style. Actually, I haven't checked if date time input exists for services in core that we could align with.

maruel:
Just as a reminder, there's already two ways to implement preheating or adjust charging level:
- Use normal home assistant time based automation 
- Use a separate Google Calendar to schedule 

I use the latter and it works great! I share the calendar with all the family members, so anyone can set the preheating. For example I'm at the gym, I know I'll leave in 1h and it's cold outside, I set a one-off event in 45min in this calendar.

Separately to the feature, I feel some documentation (read: copy pastable examples) to steer users toward the already available ways would be good idea too?

tmenguy:
@maruel sounds great the calendar part ... can you elaborate a bit on how you set it up? Thx!

InTheDaylight14:
> Separately to the feature, I feel some documentation (read: copy pastable examples) to steer users toward the already available ways would be good idea too?

The Wiki of this project is a perfect place for write-ups like that.

maruel:
I documented at https://maruel.ca/post/tesla_preheat/

InTheDaylight14:
[Wiki entry](https://github.com/alandtse/tesla/wiki/Scheduled-Charging-and-Departure) to walk-through the existing functionality wrt scheduled departure and charging. 

Right now I'm thinking the dedicated service call are not worth it. It would look very similar to the info in that Wiki entry.

tmenguy:
[...]

PR #441, opened 2022-12-18
feat: Sch departure and charging binary sensor

InTheDaylight14:


InTheDaylight14:
https://github.com/alandtse/tesla/issues/164 I don't think this fully closes the issue. We really need an input_time entity made available to integrations or the dedicated service calls.

larstobi:
> I think these binary sensors would still be helpful even if we had the full solution available. I'm on the fence with a number input being the "minutes after midnight". That's doesn't feel very intuitive.

@InTheDaylight14 

In the Tesla app, we can only set the minutes in 15 minute increments.

If I add an Entity card with the `input_datetime` helper, using only time (not datetime) then I can set hour and minute. Perhaps this helper can be used to provide input to a charger input, and it just convert the hours+minutes to only minutes before sending it to the Tesla API?

InTheDaylight14:
> In the Tesla app, we can only set the minutes in 15 minute increments.

Huh, I did not know this. Guess I have never tired to set it to odd minutes in the hour.
 
> If I add an Entity card with the `input_datetime` helper, using only time (not datetime) then I can set hour and minute. Perhaps this helper can be used to provide input to a charger input, and it just convert the hours+minutes to only minutes before sending it to the Tesla API?

HA integrations cannot make an input_datetime automatically. That entity type can only be added by individual users. My recommendation here is to use templates to convert an input_datetime into the equivalent minutes after midnight the API expects to receive. Then use that template in the custom_api call found [in the Wiki here](https://github.com/alandtse/tesla/wiki/Scheduled-Charging-and-Departure). 

If and when integrations can add a Time entity I'd like to build out switches for scheduling.

PR #448, opened 2022-12-21
feat: Add loggers to manifest for dynamic debug logging

jesserockz:
https://github.com/home-assistant/architecture/discussions/715

### DIFF
--- a/custom_components/tesla_custom/manifest.json
+++ b/custom_components/tesla_custom/manifest.json
@@ -28,5 +28,6 @@
     }
   ],
   "iot_class": "cloud_polling",
-  "version": "3.7.1"
+  "version": "3.7.1",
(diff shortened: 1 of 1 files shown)


alandtse:
Thanks. When did this feature go live? Does it support custom components? And should it include both Tesla and teslajsonpy?

jesserockz:
It does support custom components, I made the modification directly in the custom_conponents folder for tesla_custom to confirm and it worked for me.

Not sure exactly when it was added, but loggers keys were added a while back: https://github.com/home-assistant/core/pull/65083


It only needs the keys of requirements, not the actual integration itself as HA can determine that one.

https://developers.home-assistant.io/docs/creating_integration_manifest/#loggers

PR #458, opened 2022-12-23
feat: sch depart and charge timestamp attributes

InTheDaylight14:
Includes fix for scheduled departure binary state when a charger is plugged in. Discussion [here](https://github.com/alandtse/tesla/discussions/408#discussioncomment-4487304)


alandtse:
Please in the future try to separate fixes from features.

Issue #479, opened 2023-01-09
Successful service calls are not optimistic


Bre77:
When calling a service like `number.set_value` and its succesful, the state is not changed optimistically, and due to the delay in polling it can take some time before Home Assistant knows what the current value actually is.

It might be appropriate to have `async_set_native_value` actually change the internal values of **_car** if the API call was successful.

jaymunro:
I have noticed that setting the charging amps with the following results in an immediate change in current but the value displayed in HA for number.carname_charging_amps is not updated for ages. I haven't timed it but longer than 5 minutes I feel:

service: number.set_value
data:
  value: |-
    {{ 
      [0, 
        [ ( states('sensor.powerwall_excess_solar') | float  / 235 * 1000 ) | round(1)
          , ( states('input_number.car_charge_rate') | float  / 235 * 1000 ) | round(1)
        ] | min 
      ] | max
    }}
target:
  entity_id: number.carname_charging_amps

Is this the same symptom as described in this issue?

Bre77:
Yes, the value only changes on every polling interval which defaults to 660s which is 11 minutes, so 5 min is about the average.

jaymunro:
Ok, that makes sense. So I can confirm this behaviour. Though not serious, it is disconcerting when changing a value an not seeing it reflected in the UI. Makes you question the reliability of the integration.

Brett’s fix suggestion seems logical to me.

Bre77:
Ok I just checked the library, and it should already be optimistic based on this line: https://github.com/zabuldon/teslajsonpy/blob/dev/teslajsonpy/car.py#L899

I need to look into this futher as to why its not actually updating the state.

Bre77:
Ok this is an issue with the underlying library, and not the Home Assistant integration. Will be closed by https://github.com/zabuldon/teslajsonpy/pull/392 and bumping the library.

Bre77:
Can confirm using 3.7.2 fixes the number.*_charging_amps entity. Integration just needs a library bump now for others to get the fix.

alandtse:
Oops, reopening and will close once we have a real bump PR.

PR #484, opened 2023-01-15
fix: Bump teslajsonpy to fix charge current

Bre77:
A fix to the set charge current function was made in the upstream library, and so a bump is required to fix https://github.com/alandtse/tesla/issues/479

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -76,7 +76,7 @@ repos:
           - --configfile=tests/bandit.yaml
         files: ^(tests)/.+\.py$
   - repo: https://github.com/PyCQA/isort
-    rev: 5.11.4
+    rev: 5.12.0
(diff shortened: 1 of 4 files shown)


alandtse:
No need to edit the changelog as that is handled by semantic release. Please bump project.toml and update poetry.

alandtse:
Did you manually edit poetry.lock? You need to run `poetry lock` or `poetry update`. There's a hash it updates to be consisnent.

Bre77:
I'm unfamiliar with poetry and it's throwing lots of errors and trying to change multiple package versions on my Windows machine. I think I convinced it to just update the lock file content hash now.

If this still isn't correct, I'd suggest rejecting this PR and bumping using your existing development environment.

garethcook:
Still failing for me, reverting a backup

PR #534, opened 2023-03-13
fix: memorize unique id

bdraco:
slugify is a relatively expensive operation because it has to do unicode work. 

Ideally we could use `self._attr_unique_id` in `__init__` but `self.type` is set after the parent's `__init__` is called.


alandtse:
Thanks. And just FYI, the technical term is [memoize](https://en.wikipedia.org/wiki/Memoization), not memorize. ;)

PR #552, opened 2023-03-26
feat: split coordinator to avoid updating disabled cars and energy sites

bdraco:
When a user has multiple energy sites/cars they all update even if the device is disabled in Home Assistant.

Disabling a vehicle or energy site will now disable updates for it as well

needs https://github.com/zabuldon/teslajsonpy/pull/402


alandtse:
Thanks. Teslajsonpy is bumped to 3.8.0. Please remember to:

1. Update manifest.json
2. Update pyproject.toml
3. `poetry update teslajsonpy@3.8.0`

2 may be redundant of 3, but I'm not sure.

PR #569, opened 2023-04-17
feat: Add Shift State Sensor

Megabytemb:
Add a car Shift state ENUM sensor, so you can detect what gear the car is in specifically.

This is different to the Parking Break binary sensor, as you get the explicit gear.

### DIFF
--- a/custom_components/tesla_custom/sensor.py
+++ b/custom_components/tesla_custom/sensor.py
@@ -66,6 +66,7 @@ async def async_setup_entry(hass: HomeAssistant, config_entry, async_add_entitie
         entities.append(TeslaCarChargerEnergy(hass, car, coordinator))
         entities.append(TeslaCarChargerPower(hass, car, coordinator))
         entities.append(TeslaCarOdometer(hass, car, coordinator))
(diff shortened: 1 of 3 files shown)


alandtse:
Wait, isn't an enum an input class in HA?

Megabytemb:
I got it from the Sensor Dev site

https://developers.home-assistant.io/docs/core/entity/sensor/

"
The sensor has a limited set of (non-numeric) states. The options property must be set to a list of possible states when using this device class.
"

alandtse:
Ok, does our minimum version of ha cover this sensor or would we need to increment it. It's the hacs.json.

Megabytemb:
Looks like it was introduced in 2023.1.0

shall we bump the min version required? or remove the fancy enum stuff and just use a generic sensor?

alandtse:
We can bump.

alandtse:
Oh, can you please make the bump so I can merge this in?

Megabytemb:
Bumped :)

alandtse:
Thanks. But I thought you said it came in in 2023.1? The way to tell is finding the commit that brought it in.

Megabytemb:
Right you are. Fixed

PR #572, opened 2023-04-18
feat: More updates to TeslaMate

Megabytemb:
Added more values to be updates from MQTT.

Added a little extra logging

### DIFF
--- a/custom_components/tesla_custom/teslamate.py
+++ b/custom_components/tesla_custom/teslamate.py
@@ -15,17 +15,35 @@ from homeassistant.components.mqtt.subscription import (
     async_subscribe_topics,
     async_unsubscribe_topics,
 )
+from homeassistant.const import UnitOfLength
 from homeassistant.core import HomeAssistant
 from homeassistant.helpers.storage import Store
+from homeassistant.util.unit_conversion import DistanceConverter
 from teslajsonpy.car import TeslaCar
 
 from .const import TESLAMATE_STORAGE_KEY, TESLAMATE_STORAGE_VERSION
-from .util import km_to_miles
 
 if TYPE_CHECKING:
     from . import TeslaDataUpdateCoordinator
 
 logger = logging.getLogger(__name__)
+
+
+def cast_odometer(odometer: float) -> float:
+    """Convert KM to Miles.
+
+    The Tesla API natively returns the Odometer in Miles.
+    TeslaMate returns the Odometer in KMs.
+    We need to convert to Miles so the Odometer sensor calculates
+    properly.
+    """
+    odometer_km = float(odometer)
+    odometer_miles = DistanceConverter.convert(
+        odometer_km, UnitOfLength.KILOMETERS, UnitOfLength.MILES
+    )
+
+    return odometer_miles
+
 
 MAP_DRIVE_STATE = {
     "latitude": ("latitude", float),
@@ -48,7 +66,7 @@ MAP_VEHICLE_STATE = {
     "tpms_pressure_rr": ("tpms_pressure_rr", float),
     "locked": ("locked", bool),
     "sentry_mode": ("sentry_mode", bool),
-    "odometer": ("odometer", km_to_miles),
+    "odometer": ("odometer", cast_odometer),
 }
 
 MAP_CHARGE_STATE = {
--- a/custom_components/tesla_custom/teslamate.py
+++ b/custom_components/tesla_custom/teslamate.py
@@ -20,6 +20,7 @@ from homeassistant.helpers.storage import Store
 from teslajsonpy.car import TeslaCar
 
 from .const import TESLAMATE_STORAGE_KEY, TESLAMATE_STORAGE_VERSION
+from .util import km_to_miles
 
 if TYPE_CHECKING:
     from . import TeslaDataUpdateCoordinator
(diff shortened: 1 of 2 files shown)


custom_components/tesla_custom/util.py
```diff
@@ -9,3 +9,19 @@
     from homeassistant.util.ssl import client_context
 
     SSL_CONTEXT = client_context()
+
+
+def km_to_miles(odometer: float) -> float:
```
alandtse, line 14: We shouldn't have any conversion code. Is there no way to use HA native conversion code? Normally you just declare the native sensor units and HA figures it out.

custom_components/tesla_custom/teslamate.py
```diff
@@ -194,6 +213,12 @@ async def async_handle_new_data(self, car: TeslaCar, msg: ReceiveMessage):
             self.update_climate_state(car, attr, cast(msg.payload))
             coordinator.async_update_listeners()
 
+        elif mqtt_attr in MAP_CHARGE_STATE:
```
alandtse, line 216: Are there any frequent MQTT changes that we may want to explicitly mark as not being saved in the recorder?

alandtse (commented):
Please address questions.

custom_components/tesla_custom/teslamate.py
Megabytemb: Not that i've noticed. When teslaMate is connected and working, MQTT updates are about 1/Second
I would think we want the end user to be able to exclude from the recorder, compared to us.

I haven't noticed any extra obseen values in my recorder

alandtse: Sorry, are these entity attributes? I'm referring to the fact that any attribute changes will get logged in the recorder even if the main state doesn't change. This commit shows how to fix it if it does have an attribute that isn't worth recording. https://github.com/home-assistant/core/blob/3e5e937541dd368f3d70c7557ceaa7a07fd284e2/homeassistant/components/unifiprotect/recorder.py

custom_components/tesla_custom/teslamate.py
```diff
@@ -15,18 +15,36 @@
...
 from .const import TESLAMATE_STORAGE_KEY, TESLAMATE_STORAGE_VERSION
-from .util import km_to_miles
 
 if TYPE_CHECKING:
     from . import TeslaDataUpdateCoordinator
 
 logger = logging.getLogger(__name__)
 
+
+def cast_odometer(odometer: float) -> float:
```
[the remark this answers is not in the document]
Megabytemb: The issue is that the data Source has different values.
The Raw tesla API (and thus, what the integration and sensor code is expecting) is in Miles.
TeslaMate exposes the odometer in KMs.

In a weird way, by doing a conversion, we are actually keeping data consistent.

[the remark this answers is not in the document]
alandtse: I may not be explaining it well. HA sensors are now supposed to just declare whatever they natively use and HA will handle all conversions. You don't have to convert any input data anymore.

alandtse:
Also, does this cover #573

custom_components/tesla_custom/teslamate.py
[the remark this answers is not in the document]
Megabytemb: I'm Aware of the Home Assistant Conversions.

The issue here is that Teslamate introduces a second source of data.
The sensors configured in the integration expect the Raw data to be in Miles.

https://github.com/alandtse/tesla/blob/437212bb9deb3de391b6114b8fd10eac00e9b8ed/custom_components/tesla_custom/sensor.py#L282

This is because the raw data provided from the Tesla API is in Miles.

However, the MQTT data from TeslaMate is in Kilometers.
https://docs.teslamate.org/docs/integrations/mqtt/

So I need to convert the TeslaMate data back to Miles so it matches the original underlying data from the Tesla API.

Megabytemb:
> Also, does this cover #573

Latest commit fixes that issue.

I checked, all other units are the same (°C for temps, and BAR for tire pressure)

custom_components/tesla_custom/teslamate.py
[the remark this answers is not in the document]
noneabove1182: Is this not kind of expensive computationally since Teslamate is sending data every second? Would it be better to only convert Teslamate to miles if the user is using imperial, and if it's not, convert the Tesla API to metric?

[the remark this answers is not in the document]
alandtse: > I'm Aware of the Home Assistant Conversions.
> 
> The issue here is that Teslamate introduces a second source of data.
> The sensors configured in the integration expect the Raw data to be in Miles.
> 
> https://github.com/alandtse/tesla/blob/437212bb9deb3de391b6114b8fd10eac00e9b8ed/custom_components/tesla_custom/sensor.py#L282
> 
> This is because the raw data provided from the Tesla API is in Miles.
> 
> However, the MQTT data from TeslaMate is in Kilometers.
> https://docs.teslamate.org/docs/integrations/mqtt/
> 
> So I need to convert the TeslaMate data back to Miles so it matches the original underlying data from the Tesla API.
> 

Ahh yes, you're just augmenting the original sensor and HA doesn't natively support two data sources for the same sensor. Normally they'd say have two sensors and use a template sensor to normalize the data.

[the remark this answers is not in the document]
alandtse: HA is doing tons of conversions a second, I don't think that's an issue. I'm ok with using this to do the conversion.

custom_components/tesla_custom/teslamate.py
Megabytemb: So I looked into this.

Unifi is the only non-native component that uses that function (the rest are like Sensor, climate, select, etc...)
So it doesn't really seem like something we should use.

Most of the attributes we're pulling in from TeslaMate end up as States anyway, so Home Assistant's state handling should keep the database in check (aka, don't log if the value hasn't changed)

The only exception to this is
- charger_actual_current
- charger_voltage

And they are only populated While the vechile is charging, which would be getting database updates anyway due to the TeslaCarChargerPower sensor updating

In short, I don't think we need to do the recorder.py for this.

alandtse: Ok, if we're creating state sensors that's fine. The concern is that any time an attribute changes on a sensor, all the data (state and attributes) are recorded in the recorder. If something changes frequently but doesn't really mean anything in the long run, you'd want to exclude that attribute from the recorder. A contrived example might be if you also added a seconds charging attribute to a battery level. Every time the seconds changed, you'd record a new entry even if the battery state may not have changed in that one second.

I thought we were creating a lot more attributes now, so I asked. If that's not the case, then that's fine.

Megabytemb: So we're good? I don't think there's much else to add to this PR

Issue #565, opened 2023-04-05
Arrival time sensor does not update properly


stplatt:
I've been using/monitoring the Arrival time sensor for a couple of weeks and noticing that the sensor does not update as frequently as the other sensors. Based on my observation it only seems to update when the arrival time changes to a later time. I've screen grabbed some of my history and attached it so you can see what I mean. Literally every entry in my history is like this, where an arrival time only updates if it is a later time. If the arrival time updates to an earlier one (example to 6:30pm from 6:35pm), the arrival_time sensor does not update.

version 3.11.0

Standard Configuration
![arrival_time](https://user-images.githubusercontent.com/88013463/230108336-5fe4355e-7f91-4b32-b73a-6e1ebc259c0d.png)

PR #575, opened 2023-04-20
fix: Update state with earlier charge complete and arrival times

InTheDaylight14:
Added absolute value around the seconds comparison to allow for a state change when the charge complete time and the arrival time move earlier than before. 

closes https://github.com/alandtse/tesla/issues/565

### DIFF
--- a/tests/test_sensor.py
+++ b/tests/test_sensor.py
@@ -260,7 +260,7 @@ async def test_time_charge_complete_charging(
 
     assert state.state == charge_complete_str
 
-    car_mock_data.VEHICLE_DATA["charge_state"]["time_to_full_charge"] -= 0.1
+    car_mock_data.VEHICLE_DATA["charge_state"]["time_to_full_charge"] = 0.15
     earlier_charge_complete = mock_now + timedelta(
         hours=float(car_mock_data.VEHICLE_DATA["charge_state"]["time_to_full_charge"])
     )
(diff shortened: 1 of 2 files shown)


tests/test_sensor.py
```diff
@@ -260,6 +260,17 @@ async def test_time_charge_complete_charging(
...
 
+    car_mock_data.VEHICLE_DATA["charge_state"]["time_to_full_charge"] = 0.15
+    earlier_charge_complete = mock_now + timedelta(
+        hours=float(car_mock_data.VEHICLE_DATA["charge_state"]["time_to_full_charge"])
+    )
+    earlier_charge_complete_str = datetime.strftime(
+        earlier_charge_complete, "%Y-%m-%dT%H:%M:%S+00:00"
+    )
+
+    state = hass.states.get("sensor.my_model_s_time_charge_complete")
```
InTheDaylight14, line 271: This is pulling the same state from line 261. Maybe HA does not update the state between the lines or I'm not changing the correct values for HA to set a new state.

InTheDaylight14:
I've confirmed this change works in home assistant. I'm not sure how to get the state to update in the test after the times move earlier.

InTheDaylight14:
> I've confirmed this change works in home assistant. I'm not sure how to get the state to update in the test after the times move earlier.

Such a small logic change I think we are good w/o specific test

alandtse:
My preference is that we have a test to help with regressions. HA should have loads of examples of testing around mocking time. I think they use [timemachine](https://pypi.org/project/time-machine/) but there are alternatives.

InTheDaylight14:
I'm able to mock the time okay. The entity state does not update after the mocked time changes (I tired to increase and decrease the time to make sure, increasing works under the original logic). Is there a way to force HA to update the sensors?

Not working test sequence:
- Check sensor state: passes
- Change Tesla API value to arrive 3 minutes earlier than before
- Set time 2 minutes in future
- Re-check state: state has not changed from the first check. Should be 1 minute before the first check.

alandtse:
I haven't done HA testing for a while, but again, this is probably something done in core a lot.

InTheDaylight14:
I've dumped way too much time into this, I can't solve it. 

I think I've narrowed down to, the mock controller does not grab new data and update the HA states during the test. Its not an issue with the changing time, that would be easy.

alandtse:
Well, if you can't solve it just merge up to dev and I'll get you in the next release.

PR #580, opened 2023-04-25
fix: bump teslajsonpy to fix climate_state error

Megabytemb:


Megabytemb:
that seems to be errors with the build containers, and not my code?

Issue #579, opened 2023-04-21
Missing “is_auto_conditioning_on"


solvan230:
HA used to be able to expose under climate attribute “is_auto_conditioning_on" but it’s missing now. Is there anyway to bring it back?

alandtse:
Are you asking because HA removed it or the component? We won't add back stuff HA removed. If we removed it during the rewrite, it can be added back with a PR.

solvan230:
These attributes under climate disappeared when Tesla Custom Integration updated to v3.0.0

if these could be added back that would be awesome!

such as: 
    "allow_cabin_overheat_protection": true,
    "auto_seat_climate_left": true,
    "auto_seat_climate_right": false,
    "battery_heater": false,
    "battery_heater_no_power": false,
    "cabin_overheat_protection": "FanOnly",
 "cabin_overheat_protection_actively_cooling": false,
    "climate_keeper_mode": "off",
    "cop_activation_temperature": "High",
    "defrost_mode": 0,
    "driver_temp_setting": 22.8,
    "fan_status": 0,
    "hvac_auto_request": "On",
    "inside_temp": 27.0,
    "is_auto_conditioning_on": false,
    "is_climate_on": false,
    "is_front_defroster_on": false,
    "is_preconditioning": false,
    "is_rear_defroster_on": false,
    "left_temp_direction": -232,
    "max_avail_temp": 28.0,
    "min_avail_temp": 15.0,
    "outside_temp": 23.0,
    "passenger_temp_setting": 22.8,
    "remote_heater_control_enabled": false,
    "right_temp_direction": -232,
    "seat_heater_left": 0,
    "seat_heater_right": 0,
    "side_mirror_heaters": false,
    "supports_fan_only_cabin_overheat_protection": true,
    "timestamp": [1604977244530](tel:1604977244530),
    "wiper_blade_heater": false

Megabytemb:
I'm happy to work on this once i fix all the teslamate bugs. That is, if no-one gets to me first

solvan230:
I Appreciate all the good work you guys do 🙇‍♂️

PR #588, opened 2023-05-01
feat: add extra climate attributes

Megabytemb:
Adds back some attributes, as noted in https://github.com/alandtse/tesla/issues/579

### DIFF
--- a/custom_components/tesla_custom/util.py
+++ b/custom_components/tesla_custom/util.py
@@ -9,3 +9,15 @@ except ImportError:
     from homeassistant.util.ssl import client_context
 
     SSL_CONTEXT = client_context()
+
+
+def safeget(dct, *keys, default=None):
+    """Get a recursuive object from a dict."""
+    for key in keys:
+        try:
+            dct = dct[key]
+        except KeyError:
+            return default
+        except TypeError:
+            return default
+    return dct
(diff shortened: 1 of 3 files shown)


custom_components/tesla_custom/util.py
```diff
@@ -9,3 +9,15 @@
     from homeassistant.util.ssl import client_context
 
     SSL_CONTEXT = client_context()
+
+
+def safeget(dct, *keys, default=None):
```
alandtse, line 14: Since you're trying to add this in teslajsonpy, I'm not a fan of defining it twice. Let's get it in at the library level and then access it.

Megabytemb: not too keen to reuse it from teslajsonpy, as its just a helper function.
you could copy that code and run it in any python script (i use it in mine quite a bit).

if it was a bit of code specific to our tesla code, I'd agree with you, but its just a dictionary helper

alandtse: If it ends up having a bug or getting deprecated, we'll have to fix it in two places. There's no point in creating more work when you already want it in the underlying library.

jherby2k:
Hey, any chance this can get revisted? I could really use the battery heater info as cold weather approaches - looks like this got stalled by something trivial.

PR #616, opened 2023-05-26
feat: enable http/2 support when installed

bdraco:
http/2 reduces the number of TLS handshakes which has reduced the number of timeouts in testing.

This decreased the startup time from 90s to 70s for on of my larger installs.

I have a few power sites and enabling too many of them causes timeouts. Enabling h2 fixed the problem


bdraco:
There seem to be a leak in `h2` though so need to dig in a bit more to the stack.

Closing this for now and will reopen if I can solve the underlying issues

PR #622, opened 2023-06-01
fix: ensure update_vehicles coordinator always polls

bdraco:
Only create the `update_vehicles` coordinator if there are cars to be polled

Add a listener to the `update_vehicles` to ensure we continue to poll the `vehicles` endpoint since the lack of listeners previously meant the endpoint was not being polled.

### DIFF
--- a/custom_components/tesla_custom/__init__.py
+++ b/custom_components/tesla_custom/__init__.py
@@ -281,27 +281,29 @@ async def async_setup_entry(hass, config_entry):
         vins=set(),
         update_vehicles=False,
     )
+    energy_coordinators = {
+        energy_site_id: _partial_coordinator(energy_site_ids={energy_site_id})
+        for energy_site_id in energysites
+    }
+    car_coordinators = {vin: _partial_coordinator(vins={vin}) for vin in cars}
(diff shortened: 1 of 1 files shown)


custom_components/tesla_custom/__init__.py
```diff
@@ -281,14 +281,30 @@ def _async_create_close_task():
...
+        # to keep the vehicles up to date.
+        @callback
+        def _async_update_vehicles():
+            """Update vehicles coordinator.
+
+            This listener is called when the update_vehicles_coordinator
+            is updated. Since each car coordinator is also polling we don't
+            need to do anything here, but we need to have this listener
+            to ensure the update_vehicles_coordinator is updated regularly.
+            """
```
bdraco, line 305: I had this calling back the other coordinator listeners to write state but since the update interval is 10s that seemed overkill as there are already a lot of state writes

alandtse: Got it. I think it should be ok but we'll see if anyone reports issues.

alandtse:
Thanks for the quick response.

PR #626, opened 2023-06-02
fix: use async_write vs async_update

### LINKED ISSUE #606: Warning from hass when opening the charge port "self.async_update_ha_state(), without enabling force_update"

v3.12.2

## Configuration

Default config:
- 660s polling interval
- "Normal" polling policy
- Previously configured to pull from Teslamate via MQTT, disabled at the time

## Describe the bug

Opened the "more info" page for the charge port lock entity and clicked "unlock", received warning below in the logs.

## Debug log

```text
Entity lock.CAR_NAME_charge_port_latch (<class 'custom_components.tesla_custom.lock.TeslaCarChargePortLatch'>) is using self.async_update_ha_state(), without enabling force_update. Instead it should use self.async_write_ha_state(), please report it to the custom integration author.
```
I didn't enable debug logging because the message seemed pretty explicit and easily reproducible, but if necessary I can turn on debug logging and reproduce it.

InTheDaylight14:
closes #606

### DIFF
--- a/custom_components/tesla_custom/climate.py
+++ b/custom_components/tesla_custom/climate.py
@@ -128,7 +128,7 @@ class TeslaCarClimate(TeslaCarEntity, ClimateEntity):
             temp = round(temperature, 1)
 
             await self._car.set_temperature(temp)
-            await self.async_update_ha_state()
+            self.async_write_ha_state()
 
     async def async_set_hvac_mode(self, hvac_mode):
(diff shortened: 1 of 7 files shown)


alandtse:
Thanks!

alandtse:
As a note, I'll hold off on pushing this fix until the deprecation date actually comes or we have more for the release.

PR #628, opened 2023-06-12
feat: add heated steering wheel select

Megabytemb:
Adds Heated steering wheel select entity.
Disables old heat steering switch for new installs

created in Draft, as we need to wait for a new release of teslajsonpy
https://github.com/zabuldon/teslajsonpy/pull/408


Megabytemb:
alright, after a bit of GIT hell cause dumb me didn't pull from the upstream dev branch, we should be all good.

Ready for review

PR #632, opened 2023-06-13
fix: update battery remaining sensor to ENERGY_STORAGE

### LINKED ISSUE #589: HA log warning about invalid unit (Wh) for 'battery_remaining' sensor

3.12.1

## Configuration

```yaml
2023-05-01 08:56:13.225 WARNING (MainThread) [homeassistant.components.sensor] Entity sensor.jenkins_family_battery_remaining (<class 'custom_components.tesla_custom.sensor.TeslaEnergyBatteryRemaining'>) is using native unit of measurement 'Wh' which is not a valid unit for the device class ('battery') it is using; expected one of ['%']; Please update your configuration if your entity is manually configured, otherwise report it to the custom integration author.
```

## Describe the bug

Warning message in the Home Assistant log relating to an invalid unit for the sensor device class.

purcell-lab:
Fixes #589


alandtse:
Thanks. Please take care of the test failure.

alandtse:
You need to bump homeassistant to [2023.4.0](https://github.com/home-assistant/core/releases/tag/2023.4.0) in pyproject.yaml. We'll also need to update the minimum version in manifest.json to handle this.

purcell-lab:
> We'll also need to update the minimum version in manifest.json to handle this.

Sorry, I'm slowly learning my way around.

I have updated pyproject.yaml.

Do I also need to include homeassistant dependency in manifest.json?

alandtse:
Sorry, it's in the hacs.json. https://github.com/alandtse/tesla/blob/dev/hacs.json#L4  Manifest.json will break the component.  Also, you have to run poetry update so the lock file is updated properly.

purcell-lab:
> you have to run poetry update so the lock file is updated properly

Getting closer, I'm using the web interface for GitHub so unsure if I can run poetry directly.

Might be time to setup Viscose.

purcell-lab:
Thanks for pushing this through and updating the deps.

PR #657, opened 2023-07-05
feat: minutes-to-arrival-attribute

### LINKED ISSUE #656: [Feature Request] Provide active_route_minutes_to_arrival directly
**Is your feature request related to a problem? Please describe.**

See https://www.reddit.com/r/homeassistant/comments/14qepmu/comment/jqmore7/?context=3

> I'm using HACS + [Tesla Custom Integration](https://github.com/alandtse/tesla) and it worked gratefully for Charge-on-sun feature. But doing a simple automation as "When Tesla goes home, cast tts in how many time" is hard as hell.(+ when arrival time is less than 2 min, cast tts "Tesla is approaching")
> 
> Here's why:
> 1/ Tesla API give drive_state.active_route_minutes_to_arrival BUT [alandtse use it to create an arrival datetime](https://github.com/alandtse/tesla/blob/dev/custom_components/tesla_custom/sensor.py#L678C22-L678C54)
> 2/ So this value "doesn't exist" on HA side and I have to recreate it myself
> 3/ [Why the f4ck comparing datetime with offset is so hard it can fill at least a whole google page search](https://www.google.com/search?q=home+assistant+compare+two+date&rlz=1C1ONGR_frFR1064FR1064&oq=home+assistant+compare+two+date&gs_lcrp=EgZjaHJvbWUqBggAEEUYOzIGCAAQRRg70gEIMjQ3OWowajSoAgCwAgA&sourceid=chrome&ie=UTF-8)
> 4/ Comparing is not enough, I need to get minutes remaining from this given date

Provide the value of active_route_minutes_to_arrival as entity or attribute so it can be used directly in action

This one :( 
image

**Additional context**
I think I know why it has been done this way: This value is useless as it will not update itself between each tesla API call.

Maybe we could provide our own computed active_route_minutes_to_arrival ? with an event each minute and update it on each API call?

InTheDaylight14:
closes #656

### DIFF
--- a/custom_components/tesla_custom/sensor.py
+++ b/custom_components/tesla_custom/sensor.py
@@ -706,6 +706,9 @@ class TeslaCarArrivalTime(TeslaCarEntity, SensorEntity):
             "Energy at arrival": self._car.active_route_energy_at_arrival,
             "Minutes traffic delay": minutes,
             "Destination": self._car.active_route_destination,
+            "Minutes to arrival": round(
+                float(self._car.active_route_minutes_to_arrival), 2
+            ),
         }
 
 
--- a/tests/test_sensor.py
+++ b/tests/test_sensor.py
@@ -620,6 +620,10 @@ async def test_arrival_time(hass: HomeAssistant, monkeypatch: MonkeyPatch) -> No
         state.attributes.get("Destination")
         == car_mock_data.VEHICLE_DATA["drive_state"]["active_route_destination"]
     )
+    assert state.attributes.get("Minutes to arrival") == round(
+        car_mock_data.VEHICLE_DATA["drive_state"]["active_route_minutes_to_arrival"],
+        2,
+    )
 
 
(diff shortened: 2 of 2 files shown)


alandtse:
Wait, didn't we already have arrival time eta?

InTheDaylight14:
Yes, in the format of a timestamp. This is adding the raw "minutes to arrival" from the api and to the attributes of the timestamp sensor as requested in the issue. This will make it easier for ppl to trigger automations on the arrival time dropping below 5 min or similar. There is a workaround posted in the issue for a custom template sensor that does the same thing.

alandtse:
The sensor seems redundant. Would it solve the problem by just doing the attribute only? HA basically requires timers to be timestamps. I don't think we should be encouraging an alternative paradigm.

storm1er:
I think the attribute would be enough, no need to create a full entity for this.

PR #670, opened 2023-07-21
fix: small cleanups to entity construction

bdraco:
- Reduce some lookups when writing state
- Remove `__init__` delegations where the attributes are static

### DIFF
--- a/custom_components/tesla_custom/base.py
+++ b/custom_components/tesla_custom/base.py
@@ -1,5 +1,5 @@
 """Support for Tesla cars and energy sites."""
-from homeassistant.core import HomeAssistant
+from homeassistant.core import HomeAssistant, callback
 from homeassistant.helpers.entity import DeviceInfo
 from homeassistant.helpers.update_coordinator import CoordinatorEntity
 from homeassistant.util import slugify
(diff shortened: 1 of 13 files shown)


custom_components/tesla_custom/base.py
```diff
@@ -17,22 +17,23 @@ class TeslaBaseEntity(CoordinatorEntity):
...
     def __init__(
         self, hass: HomeAssistant, coordinator: TeslaDataUpdateCoordinator
     ) -> None:
         """Initialise the Tesla device."""
         super().__init__(coordinator)
         self._coordinator: TeslaDataUpdateCoordinator = coordinator
-        self._enabled_by_default: bool = True
         self.hass = hass
-        self.type = None
         self._memorized_unique_id = None
```
bdraco, line 30: This could likely be cleaned up more since `CoordinatorEntity` takes care of setting `self.coordinator` and `self.hass` is set in `async_added_to_hass` but that would require more invasive refactoring and I didn't want to make this larger

alandtse: No worries. Thanks for the cleanup you did do.

bdraco: Thanks. Did this in https://github.com/alandtse/tesla/pull/671

Also finally had some time to look into why startup was so slow when the car is asleep https://github.com/zabuldon/teslajsonpy/pull/416

PR #672, opened 2023-07-25
refactor: reduce code needed to construct entities

bdraco:


alandtse:
Thanks. Just to confirm, you've marked this as fix. But does it actually solve a bug or issue? Or was it just to clean up code.

bdraco:
Its just code cleanup. Intent was to generate a release

alandtse:
I think our ci settings only generate a release on feat or bug. I'll see if I can tweak it for perf and refactor too. Didn't you solve a wake issue? Pulling in the new dependency may count as a fix.

bdraco:
> I think our ci settings only generate a release on feat or bug. I'll see if I can tweak it for perf and refactor too. Didn't you solve a wake issue? 


> Pulling in the new dependency may count as a fix. 


https://pypi.org/project/teslajsonpy/

It looks like the publish failed

alandtse:
Thanks. Fixed the pypi publish. 3.9.2

PR #673, opened 2023-07-27
fix: add missing CONFIG_SCHEMA and service translations

bdraco:
BREAKING CHANGE: Home Assistant Core 2023.6.0 or later is now required

fixes https://github.com/alandtse/tesla/actions/runs/5677167144/job/15385034940

Note: we can't remove from services.yaml since old version still need it

### DIFF
--- a/custom_components/tesla_custom/__init__.py
+++ b/custom_components/tesla_custom/__init__.py
@@ -17,6 +17,7 @@ from homeassistant.const import (
 )
 from homeassistant.core import callback
 from homeassistant.exceptions import ConfigEntryAuthFailed, ConfigEntryNotReady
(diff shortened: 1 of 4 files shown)


alandtse:
Thanks for fixing this. Since this is a really recent HA release, I may hold this for a release or too.

Olen:
The missing service names causes other PRs to fail validation, as HASSFEST requires them:
https://github.com/alandtse/tesla/actions/runs/5856579213/job/15876614220?pr=685

alandtse:
Thanks; merged. I'll plan a release for this in the next month or so unless we see something like the beta causing issues or we have another need to release.

PR #674, opened 2023-07-27
fix: bump teslajsonpy to 3.9.2

bdraco:
changelog: https://github.com/zabuldon/teslajsonpy/compare/v3.9.0...v3.9.2

fixes slow startup, fixes energy sites sometimes not coming available, improves performance


bdraco:
thanks

PR #698, opened 2023-08-27
fix: Improve a couple of things in MQTT sync from Teslamate

### LINKED ISSUE #689: MQTT boolean conversion issue
## Configuration
Teslamate MQTT

## Describe the bug
Boolean entities are not updated correctly from MQTT (e.g. sentry mode, is_climate_on). I think it's due to ReceiveMessage returning msg.payload as a string, which is then casted to bool. But this doesn't work since bool("false") or even bool("False") always returns True.

cguedel:
Fixes #689

- parse boolean strings for is_climate_on, locked, sentry_mode
- add sync for plugged_in
- add sync for charge_port_door_open
- add sync for trunk_open
- add sync for frunk_open
- add sync for is_user_present
- add sync for is_preconditioning

### DIFF
--- a/custom_components/tesla_custom/teslamate.py
+++ b/custom_components/tesla_custom/teslamate.py
@@ -46,16 +46,16 @@ def cast_odometer(odometer: float) -> float:
 
 
 def cast_plugged_in(val: str) -> str:
-    """Convert boolean string for plugged_in value"""
+    """Convert boolean string for plugged_in value."""
     return "Connected" if cast_bool(val) else "Disconnected"
 
 
 def cast_bool(val: str) -> bool:
-    """Convert bool string to actual bool"""
+    """Convert bool string to actual bool."""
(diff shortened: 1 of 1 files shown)


alandtse:
Thanks. I think the test failure may be unrelated but there's some lint issues. Can you please address?

cguedel:
Should be ok now I think, no idea about the tests though

oakedcorn:
Thanks for the fix!
There's a typo in the code here:

```
def cast_trunk_open(val: str) -> int:
    """Convert bool string to trunk/frunk open/close value."""
    return 255 if cast_bool(str) else 0
```

cast_bool(str) should be cast_bool(val)

cguedel:
Sorry about that, I'll take care of today

oakedcorn:
No worries, thanks for supporting this great integration!

PR #715, opened 2023-09-18
feat: Add MQTT sync from Teslamate for battery_range

dagstuan:
Re-use cast-method from odometer since they both are reported in km from Teslamate and need to be converted to miles for teslajsonpy to be happy.


dagstuan:
@alandtse any chance of getting this merged? 🙏

PR #734, opened 2023-10-08
feat: Sync car state from Teslamate MQTT

### LINKED ISSUE #712: Sync MQTT charging state via Teslamate
Teslamate exposes an event when the Tesla starts to charge via `teslamate/cars/$car_id/state` and the payload is `charging` but this is not synced with your plugin `binary_sensor.carname_charging` entity. It would be nice to have this synced.

dagstuan:
Since Teslamate doesn't map directly to TeslaCar from teslajsonpy,
we need to have special handling of both 'plugged_in' and 'state'
when they come from Teslamate. When we receive a new 'plugged_in'-value,
we check the `car.state` to determine the new 'charging_state'
in teslajsonpy, and similarly we need to check the `car.charging_state`
when Teslamate sends a new car-state.

Also refactor the MQTT-code slightly to avoid multiple implementations
of the code that updates the TeslaCar-object.

Fixes #712

### DIFF
--- a/custom_components/tesla_custom/number.py
+++ b/custom_components/tesla_custom/number.py
@@ -56,12 +56,20 @@ class TeslaCarChargeLimit(TeslaCarEntity, NumberEntity):
     @property
     def native_min_value(self) -> int:
         """Return min charge limit."""
-        return self._car.charge_limit_soc_min
+        return (
+            self._car.charge_limit_soc_min
+            if self._car.charge_limit_soc_min is not None
+            else 0
+        )
 
     @property
     def native_max_value(self) -> int:
         """Return max charge limit."""
-        return self._car.charge_limit_soc_max
+        return (
+            self._car.charge_limit_soc_max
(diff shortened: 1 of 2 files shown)


alandtse:
FYI, not planning to merge any feature PRs until the tests are cleaned up. So it may be a long wait since people almost never contribute to fix tests.

dagstuan:
> FYI, not planning to merge any feature PRs until the tests are cleaned up. So it may be a long wait since people almost never contribute to fix tests.

I see. Do you know how the tests are broken? If you submit a commit fixing one of them I could probably help fixing some of them, they all look like they fail on roughly the same error.

alandtse:
Are you not able to see the broken tests? It should show if you look at the failing CI.

dagstuan:
> Are you not able to see the broken tests? It should show if you look at the failing CI.

I can see them, yes. But I'm not well-versed in how to go about fixing them. This PR doesn't fail any new ones does it?

Issue #737, opened 2023-10-10
FR: Update service call to return JSON data response


purcell-lab:
**Is your feature request related to a problem? Please describe.**

Home Assistant Services can now respond with data:
https://www.home-assistant.io/blog/2023/07/05/release-20237/#services-can-now-respond
The technical developer reference is available here:
https://developers.home-assistant.io/docs/dev_101_services/#response-data

The Tesla Custom Integration provides for a the calling of services in particular the calling of end points, but currently doesn't provide the data response associated with a call:

![image](https://github.com/alandtse/tesla/assets/79175134/11d2f2fb-c9d0-4632-b2e5-da596e3c85cb)

**Describe the solution you'd like**
It would be great for debugging and creating additional sensors if the custom integration service could be updated to provide the JSON response:

```json
cli.py -e mark@purcell.id.au -z PRODUCT_LIST
{
    "response": [
        {
            "id": XXX,
            "user_id": XXX,
            "vehicle_id": XXX,
            "vin": "XXX",
            "color": null,
            "access_type": "DRIVER",
            "display_name": "Betty",
            "option_codes": null,
```

**Describe alternatives you've considered**
I am currently using additional python scripts to query the API and provide the JSON responses for the additional end points I am interested in.

```
$ ./get_period.py 
{"tou_period": "SUPER_OFF_PEAK", "buy_cost": 0.01, "sell_price": 0}

#!/usr/bin/env python3

try:
    import teslapy
except ImportError:
    print("Error: The teslapy module is not installed.")
    exit()

import teslapy
import datetime
import json

def get_current_tou_period(seasons_data):
    current_hour = datetime.datetime.now().hour
    current_minute = datetime.datetime.now().minute
    current_day_of_week = datetime.datetime.now().weekday()

    for season, data in seasons_data.items():
        for period, time_ranges in data['tou_periods'].items():
            for time_range in time_ranges:
                from_hour = time_range['fromHour']
                to_hour = time_range['toHour']
                from_minute = time_range['fromMinute']
                to_minute = time_range['toMinute']
                from_day_of_week = time_range['fromDayOfWeek']
                to_day_of_week = time_range['toDayOfWeek']

                if from_day_of_week <= current_day_of_week <= to_day_of_week:
                    if from_hour < to_hour: 
                        if (from_hour < current_hour < to_hour) or (from_hour == current_hour and from_minute <= current_minute) or (to_hour == current_hour and current_minute < to_minute):
                            return period
[...]

PR #750, opened 2023-10-26
feat: Update services.py to return responses

purcell-lab:
Fixes https://github.com/alandtse/tesla/issues/737


alandtse:
Thanks. Until the tests are fixed, all PRs are on hold for now.  So there may be a delay until someone decides to fix them.

PR #753, opened 2023-10-28
test: Fix tests

functionpointer:
Fix the tests!

Most tests just had a broken `hass.services.async_call`.
`hass.services.async_call` no longer returns a boolean, as documented here:
https://developers.home-assistant.io/blog/2023/06/14/service-calls/
Fix: just don't look at the return value anymore.

However, some tests fail due to mismatch in mock data and teslajsonpy and/or extraction code in `sensor.py`.
I have fixed them by modifying the mock data.

Not sure if that is the correct approach. Are there recent examples of SolarPowerwallSite responses?


alandtse:
Thanks for doing that. Modifying mock is fine as long as they actually logically make sense still as a test.

Issue #766, opened 2023-11-09
TeslaMate provided location bounces to "unknown"


k-korn:
## Version of the custom_component
latest, 3.19.1

## Configuration

Configured via UI.

## Describe the bug

Due to recent changes in Tesla API, they stopped providing car location. Instead, I've switched to TeslaMate MQTT approach with coordinates posted to teslamate/cars/1/latitude, etc. 

In general, this approach works. However, every few minutes car location changes to "unknown" and then bounces back:
https://i.imgur.com/ykkws8F.png
Events at 17:05:51 and 17:06:12 from the screenshot are visible in the debug log below.

If I check HA/Developer Tools/Events and subscribe to "state_changed" event, the following can be seen:
````
event_type: state_changed
data:
  entity_id: device_tracker.model_3_location_tracker
  old_state:
    entity_id: device_tracker.model_3_location_tracker
    state: Parking
    attributes:
      source_type: gps
      latitude: 30.453672
      longitude: 23.238591
      gps_accuracy: 0
      heading: null
      speed: null
      attribution: Data provided by Tesla
      friendly_name: Model 3
      icon: mdi:car-electric
    last_changed: "2023-11-09T14:54:12.564300+00:00"
    last_updated: "2023-11-09T14:54:12.564300+00:00"
    context:
      id: 01HET9752MDXNFSTYNNAQHB7MG
      parent_id: null
      user_id: null
  new_state:
    entity_id: device_tracker.model_3_location_tracker
    state: unknown
    attributes:
      source_type: gps
      heading: null
      speed: null
      attribution: Data provided by Tesla
      friendly_name: Model 3
      icon: mdi:car-electric
    last_changed: "2023-11-09T14:56:53.098924+00:00"
    last_updated: "2023-11-09T14:56:53.098924+00:00"
    context:
      id: 01HET9C1VADAMNTG24D73SMHSN
      parent_id: null
      user_id: null
origin: LOCAL
time_fired: "2023-11-09T14:56:53.098924+00:00"
context:
  id: 01HET9C1VADAMNTG24D73SMHSN
  parent_id: null
  user_id: null

````
In the "new_state" the state itself is "unknown" and the coordinates are missing, probably causing the "unknown" state.
This issue is most likely related to https://github.com/alandtse/tesla/pull/608, but I'm not sure if debounce is tunable. 

## Debug log

```
2023-11-09 17:04:42.589 DEBUG (MainThread) [custom_components.tesla_custom] Listeners updated
[...]

PR #770, opened 2023-11-11
fix: Bump teslajsonpy to 3.9.7

### LINKED ISSUE #756: Beginning with 2023.38, location tracking fails
Beginning with 2023.38, Tesla introduced controls on location sharing. Location information now requires an approval grant from within the vehicle.

Relevant information can be found in the [vehicle_data endpoint](https://developer.tesla.com/docs/fleet-api#vehicle) in the Fleet API doc:

_For vehicles running firmware versions 2023.38+, location_data is required to fetch vehicle location. This will result in a location sharing icon to show on the vehicle UI._

dagstuan:
Supply 'endpoints'-parameter to 'vehicle_data' endpoint which fixes the missing location data after upgrading vehicles to 2023.38.*

The changes in teslajsonpy should probably be tested some more, I have not tested that this doesn't break any existing logic.

Should hopefully fix #756 and #766


alandtse:
Poetry also needs to be updated. See other commits with bumps.

Zappo-II:
FYI: Just bumping teslajsonpy did not fix it (#756) for me, seems like even more sensors besides location data get broken then...

dagstuan:
Okay @alandtse I attempted to upgrade for Poetry as well. Though I have never used it before so I might have done something wrong.

dagstuan:
I noticed while I was driving that the sensors in HA would become "unavailable". Turns out that without URL encoding the semicolons the Tesla API sometimes do not return all of the requested data. So this PR in teslajsonpy will probably need to be merged first: https://github.com/zabuldon/teslajsonpy/pull/437

PR #772, opened 2023-11-12
fix: Bump teslajsonpy to 3.9.8

Kokahontas13:
In order to fix issue #756, we need to have https://github.com/zabuldon/teslajsonpy/pull/437 merged. This was implemented in teslajsonpy version 3.9.8. As mentioned by [dagstuan](https://github.com/dagstuan) in #770

### DIFF
--- a/custom_components/tesla_custom/manifest.json
+++ b/custom_components/tesla_custom/manifest.json
@@ -23,6 +23,6 @@
   "iot_class": "cloud_polling",
   "issue_tracker": "https://github.com/alandtse/tesla/issues",
   "loggers": ["teslajsonpy"],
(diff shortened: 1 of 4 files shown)


Kokahontas13:
Sensors and location tracker are working fine with version 3.9.8

alandtse:
Please also run `poetry update` so the lock file is synced.

Kokahontas13:
updated

alandtse:
Thanks

BogdanDIA:
Right, was trying last night to add those endpoints but they did not work so discovered that semicolon needs to be escaped.
Thanks for the fix. It works now with the changes in teslajsonpy 3.9,8

PR #803, opened 2023-12-23
fix: calculate assumed state only once

bdraco:
assumed_state on the entities was calculated for each entity. We can calculate it once per update instead.


alandtse:
Sounds like this may be a perf: since it could improve performance.

bdraco:
Thanks

alandtse:
Thanks for the contrib. My biggest concern is we may lose all api access in a few days.

PR #806, opened 2023-12-26
fix: climate platform with HA 2024.1.x

bdraco:


alandtse (commented):
Should we also set a HA minimum in hacs.json?

bdraco:
2023.7.x is more than new enough.  This is a missed deprecation that will finally break in 2024.1.x (unless it gets extended again)

bdraco:
thanks

Issue #810, opened 2024-01-05
Deprecated constant in HA 2025.1


sh4rks:
## Version of the custom_component

Latest (3.19.5)

## Describe the bug

Some constants will disappear in HA 2025.1

## Debug log

Logger: homeassistant.const
Source: helpers/deprecation.py:204
First occurred: 4 janvier 2024 à 16:31:40 (14 occurrences)
Last logged: 4 janvier 2024 à 16:32:06

PRESSURE_BAR was used from tesla_custom, this is a deprecated constant which will be removed in HA Core 2025.1. Use UnitOfPressure.BAR instead, please create a bug report at https://github.com/alandtse/tesla/issues
PRESSURE_PSI was used from tesla_custom, this is a deprecated constant which will be removed in HA Core 2025.1. Use UnitOfPressure.PSI instead, please create a bug report at https://github.com/alandtse/tesla/issues
SPEED_MILES_PER_HOUR was used from tesla_custom, this is a deprecated constant which will be removed in HA Core 2025.1. Use UnitOfSpeed.MILES_PER_HOUR instead, please create a bug report at https://github.com/alandtse/tesla/issues
TEMP_CELSIUS was used from tesla_custom, this is a deprecated constant which will be removed in HA Core 2025.1. Use UnitOfTemperature.CELSIUS instead, please create a bug report at https://github.com/alandtse/tesla/issues
ELECTRIC_CURRENT_AMPERE was used from tesla_custom, this is a deprecated constant which will be removed in HA Core 2025.1. Use UnitOfElectricCurrent.AMPERE instead, please create a bug report at https://github.com/alandtse/tesla/issues

Logger: homeassistant.components.device_tracker
Source: helpers/deprecation.py:204
Integration: Dispositif de suivi (documentation, issues)
First occurred: 4 janvier 2024 à 16:32:06 (2 occurrences)
Last logged: 4 janvier 2024 à 16:32:06

SOURCE_TYPE_GPS was used from tesla_custom, this is a deprecated constant which will be removed in HA Core 2025.1. Use SourceType.GPS instead, please create a bug report at https://github.com/alandtse/tesla/issues

alandtse:
Thanks for reporting. We're near the end of the lifecycle for this, but perhaps we can resolve it.

ZoomeoTooknor:
It seems that no action is functional anymore (opening windows, managing temperature, heating, etc.)

Sandvoxel:
[...]

PR #814, opened 2024-01-09
fix: remove use of deprecated HA constants

epenet:
Fixes https://github.com/alandtse/tesla/issues/810


alandtse:
Thanks. I may hold this back for a little since the deprecation is in a year. We probably need to bump the HA version too to whenever the new constants came in.

The alternative may be to do a exception catch for ImportError and fallback to the old one. Then there wouldn't need to be a HA min bump.

epenet:
I think all the constants are available since 2022... They just never raised warnings before.
See https://developers.home-assistant.io/blog/2023/12/19/constant-deprecation

So we don't need to bump the HA version.

Regarding the CI - I'm not sure why it's failing now. Maybe a temporary error on Github?

alandtse:
Yah CI is a temporary error. If you are sure these don't require a min HA version, then I'm ok with it.

PR #868, opened 2024-02-23
fix: Trunk not closing

SmarterHomeLife:
Add open and close to supported features attributes and remove check for powered lift gate from trunk section to fix trunk not closing in release 3.19.8 and above.

### DIFF
--- a/custom_components/tesla_custom/cover.py
+++ b/custom_components/tesla_custom/cover.py
@@ -101,7 +101,7 @@ class TeslaCarTrunk(TeslaCarEntity, CoverEntity):
     _attr_device_class = CoverDeviceClass.DOOR
     _attr_icon = "mdi:car-back"
     _attr_supported_features = CoverEntityFeature.OPEN | CoverEntityFeature.CLOSE
-    
+
     async def async_close_cover(self, **kwargs):
         """Send close cover command."""
         _LOGGER.debug("Closing cover: %s", self.name)
--- a/custom_components/tesla_custom/cover.py
+++ b/custom_components/tesla_custom/cover.py
@@ -100,7 +100,8 @@ class TeslaCarTrunk(TeslaCarEntity, CoverEntity):
     type = "trunk"
     _attr_device_class = CoverDeviceClass.DOOR
     _attr_icon = "mdi:car-back"
-
+    _attr_supported_features = CoverEntityFeature.OPEN | CoverEntityFeature.CLOSE
+    
     async def async_close_cover(self, **kwargs):
         """Send close cover command."""
         _LOGGER.debug("Closing cover: %s", self.name)
@@ -119,14 +120,6 @@ class TeslaCarTrunk(TeslaCarEntity, CoverEntity):
     def is_closed(self):
         """Return True if trunk is closed."""
         return self._car.is_trunk_closed
-
-    @property
-    def supported_features(self) -> int:
-        """Return supported features."""
-        if self._car.powered_lift_gate:
-            return CoverEntityFeature.OPEN | CoverEntityFeature.CLOSE
-
-        return CoverEntityFeature.OPEN
 
 
 class TeslaCarWindows(TeslaCarEntity, CoverEntity):
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -51,7 +51,7 @@ repos:
   - hooks:
       - id: black
     repo: https://github.com/psf/black
-    rev: 24.1.0
+    rev: 24.2.0
   - repo: https://github.com/pre-commit/mirrors-prettier
     hooks:
       - id: prettier
--- a/custom_components/tesla_custom/const.py
+++ b/custom_components/tesla_custom/const.py
@@ -1,6 +1,6 @@
 """Const file for Tesla cars."""
 
-VERSION = "3.19.8"
+VERSION = "3.19.9"
 CONF_EXPIRATION = "expiration"
 CONF_INCLUDE_VEHICLES = "include_vehicles"
 CONF_INCLUDE_ENERGYSITES = "include_energysites"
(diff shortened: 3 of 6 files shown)


custom_components/tesla_custom/cover.py
```diff
@@ -120,14 +121,6 @@ def is_closed(self):
         """Return True if trunk is closed."""
         return self._car.is_trunk_closed
 
-    @property
-    def supported_features(self) -> int:
-        """Return supported features."""
-        if self._car.powered_lift_gate:
```
alandtse, line 126: Are we sure that this value is no longer reliable? This is coming from the teslajsonpy api and may need to be fixed there.

SmarterHomeLife: It doesn't appear to be in 3.19.8 and above, also I'm not aware of any Tesla vehicle that does not have a powered trunk lift.

alandtse: Older Model 3s and Ss do not have power trunks.

SmarterHomeLife: Apologies I didn't know that.

alandtse: Enable debug logging and it should start spitting out the API values from teslajsonpy.

Apparently it was under the "plg" key. 
https://github.com/zabuldon/teslajsonpy/blob/94f478d4819fc9887919e899bbff48eec6045ea7/teslajsonpy/car.py#L454

You can determine if it's still there or if it's a new value now with the API change.

alandtse: Odd, if you back out your change does it stop working? It used to be under the section `vehicle_config`. Maybe it moved?

SmarterHomeLife: Yes if I back out my change it stops.  It looks to still be in the ```vehicle_config``` section in the debug log.

alandtse: Thanks to be honest, I'm not thrilled with fixing this by disabling the check. I'll leave this open but my preferred fix is to actually find out why the setting isn't being detected properly since the API is reporting properly.

SmarterHomeLife: Totally agree, as I said I wasn't aware there were older vehicles without a powered lift gate so it's important that the integration continues to work for everyone.  Is the ```vehicle_config``` section the only place this is reported, is it possible that something in the vehicle details coming from Tesla is overriding it?

alandtse: I think someone would have to debug how we're updating `self._car.powered_lift_gate`. You're not using the Teslamate MQTT integration to pull in data right? I remember there was some bounciness in values when that was pulling in info too.

SmarterHomeLife: No I'm not using Teslamate.

artem-zinnatullin: Just a stupid thought: could we somehow use car VIN (model year, etc) or other property to nail down the cars with power trunk? 

We're so used to the car opening and closing trunk when we get home and it's not working for quite some time now, I'm also not using Teslamate

kbrint: This is a pragmatic idea!  The 10th digit of the VIN indicates the model year.

https://www.alldata.com/us/en/support/repair-collision/article/vin-to-year-chart

kbrint: Also, what's the harm in allowing attempted close, if the car doesn't support it?  I think the HA integration should expose a close option even if not all cars support it.

I suggest that exposing a capability that the car DOES NOT have is strictly better than NOT exposing the capability on cars that DO support it.  The fact that the majority of cars DO support it only reinforces the point.

alandtse: I'm not going to introduce a fix that removes functionality that used to be there unless it's obviously broken. Right now, the data is supposedly there and I'm waiting for someone to debug what has changed. Once we know that, then we can decide on whether we ignore this requirement.

And yes, it's problematic to assume everyone wants it. I guarantee there will be people complaining that the `close` doesn't work and someone will have to explain they need a power lift gate.

Avatar1976: > I'm not going to introduce a fix that removes functionality that used to be there unless it's obviously broken. Right now, the data is supposedly there and I'm waiting for someone to debug what has changed. Once we know that, then we can decide on whether we ignore this requirement.
> 
> And yes, it's problematic to assume everyone wants it. I guarantee there will be people complaining that the `close` doesn't work and someone will have to explain they need a power lift gate.

How would someone debug self._car.powered_lift_gate ?

kbrint: I enabled logging and did some poking around.  Findings so far:
 * The JSON response to the `VEHICLE_DATA` query _does_ have the `vehicle_config` dict, and the value of `plg` inside that dict is correct for my car.  The JSON looks good.
 * The `TeslaCar` object (stored in the custom component as `self._car`) stores some but not all of the information from the JSON response.
 * T`self._car._car.keys()` contains only:  `dict_keys(['id', 'user_id', 'vehicle_id', 'vin', 'color', 'access_type', 'display_name', 'option_codes', 'cached_data', 'granular_access', 'tokens', 'state', 'in_service', 'id_s', 'calendar_enabled', 'api_version', 'backseat_token', 'backseat_token_updated_at', 'ble_autopair_enrolled', 'command_signing'])`
   * Some of these keys (e.g. `command_signing`, `ble_autopair_enabled`) are only present in the `PRODUCT_LIST` query.
   * teslajsonpy claims that it's "Update VEHICLE_DATA", but the `TeslaCar` object does not get any new keys afterwards.

Therefore, I think this is NOT a problem with the custom component, but rather it's a problem with teslajsonpy.

alandtse:
Covered by #932. Thanks everyone particularly @kbrint @SmarterHomeLife for working on the problem.

PR #875, opened 2024-03-02
fix: avoid blocking the event loop during import

bdraco:
This option is new for 2024.3. Previously all integrations had to be imported in the event loop which would block the loop while it happened.

This option is safe to add for pre 2024.3 as it will do nothing.

This could be considered `pref` but not blocking the event loop is might always considered a fix

### DIFF
--- a/custom_components/tesla_custom/manifest.json
+++ b/custom_components/tesla_custom/manifest.json
@@ -22,8 +22,8 @@
   "documentation": "https://github.com/alandtse/tesla/wiki",
   "iot_class": "cloud_polling",
   "issue_tracker": "https://github.com/alandtse/tesla/issues",
+  "import_executor": true,
   "loggers": ["teslajsonpy"],
   "requirements": ["teslajsonpy==3.9.11"],
-  "import_executor": true,
   "version": "3.19.9"
 }
--- a/custom_components/tesla_custom/manifest.json
+++ b/custom_components/tesla_custom/manifest.json
@@ -20,8 +20,8 @@
     }
   ],
   "documentation": "https://github.com/alandtse/tesla/wiki",
-  "iot_class": "cloud_polling",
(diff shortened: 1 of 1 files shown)


custom_components/tesla_custom/manifest.json
```diff
@@ -22,6 +22,7 @@
   "documentation": "https://github.com/alandtse/tesla/wiki",
   "iot_class": "cloud_polling",
   "issue_tracker": "https://github.com/alandtse/tesla/issues",
```
bdraco, line 24:
```suggestion
  "import_executor": true,
  "issue_tracker": "https://github.com/alandtse/tesla/issues",
```

custom_components/tesla_custom/manifest.json
```diff
@@ -22,6 +22,7 @@
...
+  "import_executor": true,
```
bdraco, line 25:
```suggestion
```

alandtse:
Thanks. This looks good to me. Please note the hassfest error due to the later keys needing to be alphabetical. I'm hoping to do a release this weekend. Waiting on #853 and this.

custom_components/tesla_custom/manifest.json
```diff
@@ -21,6 +21,7 @@
   ],
   "documentation": "https://github.com/alandtse/tesla/wiki",
   "iot_class": "cloud_polling",
```
bdraco, line 23:
```suggestion
  "import_executor": true,
  "iot_class": "cloud_polling",
```

custom_components/tesla_custom/manifest.json
```diff
@@ -21,6 +21,7 @@
...
+  "import_executor": true,
```
bdraco, line 24:
```suggestion
```

PR #881, opened 2024-03-06
fix: check sensor value is valid before casting

### LINKED ISSUE #880: Battery status not available after last update
after the last update the integration no longer calls the battery status "This entity is no longer provided by the tesla_custom integration. If the entity is no longer used, delete it in settings." happened to anyone else?

alandtse:
closes #880


bdraco:
confirmed fix

PR #890, opened 2024-03-12
fix: key error for CONF_API_PROXY_CERT

### LINKED ISSUE #889: Key Error with vehicles that do not use the TESLA proxy (on version 3.20)
My apologies but I lost the logs with the exact details of the error after I downgraded to v 3.19.11.   However when I tried v3.20.00 and did not install the API proxy (since I own a 2018 Model X) I got a 'key error' on the following line (line 144 - custom_components/tesla_custom/__init__.py) :

if config[CONF_API_PROXY_CERT]:

llamafilm:
Fixes #889


alandtse:
Thanks!

PR #892, opened 2024-03-12
build: Bump dev to Home Assistant 2023 and python 3.12

### LINKED ISSUE #876: Update dev environment for Python 3.12
Home Assistant deprecated support for Python 3.10 a few months ago in https://github.com/home-assistant/core/pull/97007 and the core container currently runs 3.12.1.  I suggest updating this devcontainer environment to match.  I was going to include this as part of my existing PR, but I realized it may have some implications for dependencies so I'll keep it separate.

I'm not sure if this is related, but I get several ModuleNotFound exceptions when trying to start this after rebuilding the dev container.

llamafilm:
This just updates the dev environment to make a more accurate real world test.  I haven't tested this in real life yet, I only ran it for a few minutes in the dev container and verified that everything comes up fine.

Closes #876.

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -1,6 +1,6 @@
 {
   "name": "Tesla Custom Component",
-  "image": "mcr.microsoft.com/vscode/devcontainers/python:0-3.10-bullseye",
+  "image": "mcr.microsoft.com/devcontainers/python:1-3.12",
   "postCreateCommand": "sudo scripts/setup",
   "appPort": ["9123:8123"],
   "customizations": {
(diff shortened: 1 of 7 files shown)


llamafilm:
Unfortunately prospector doesn't yet support 3.12 so the pre-commit hook fails.
We need to update pylint to v3.0.0 or higher, but prospector is not compatible.  Looks like this PR will fix it: https://github.com/landscapeio/prospector/pull/643

alandtse:
Prospector has always given a bit of trouble so maybe we find a replacement if that doesn't get merged soon.

PR #909, opened 2024-03-14
fix: set data to default of None

### LINKED ISSUE #907: `return data` statement code path where data is not set

HEAD as of now; 1c59bc28e2091737998906d52db38c87d488c5c7

## Describe the bug

https://github.com/alandtse/tesla/blob/1c59bc28e2091737998906d52db38c87d488c5c7/custom_components/tesla_custom/__init__.py#L484

data is set on line 461 but if IncompleteCredentials triggers, it does a reload and still return data that is not existing. I think a return after lin 474 would fix the exception.

## Debug log

```text
2024-03-13 21:29:00.273 ERROR (MainThread) [custom_components.tesla_custom] Unexpected error fetching tesla_custom data: cannot access local variable 'data' where it is not associated with a value     
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 313, in _async_refresh                                                                                                 
    self.data = await self._async_update_data()                                                                                                                                                          
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                          
  File "/config/custom_components/tesla_custom/__init__.py", line 484, in _async_update_data                                                                                                                 return data                                                                                                                                                                                          
           ^^^^                                                                                                                                                                                          UnboundLocalError: cannot access local variable 'data' where it is not associated with a value 
```

maruel:
Fixes #907


alandtse:
I haven't tested this, did you actually run this to see if it helps? I believe it's harmless enough with the existing code though but would appreciate confirmation.

maruel:
I haven't, I just read the code and found the bug. To reproduce, one has to have expired credentials, to get in the except clause on line 467 and without reload_lock held. I believe this happens if you get there by clicking on the notification. Then line 485 blews up because data is undefined via this code path. This PR just makes sure data is always defined.

PR #942, opened 2024-04-13
fix: bump teslajsonpy to 3.10.3 (#919)

gkwok1:


alandtse:
Thanks. Are you able to run poetry update so it syncs the lock file?

alandtse:
Unfortunately whatever fixes were included here appear to have broken the battery tests. Those will need to be fixed before it gets merged.

gkwok1:
Removed the energy_left from the test data and updated the data calculation. Should be good to now.

alandtse:
I mean unless the test isn't needed anymore because it's not in the API, removing a failing test isn't the way to address it. Can you please confirm you removed it because it's not in the API?

gkwok1:
I removed the energy left elements from the mock data and kept the test. The test is now using the same formula as what is used teslajsonpy.

Issue #946, opened 2024-04-22
Dynamic Polling Automation not working


ddaddy:
I am setting up the Dynamic Polling as explained here: https://github.com/alandtse/tesla/wiki/Dynamic-Polling-Frequency

When I run the action of an automation to set it to 30s I see no difference.

What exactly is the binary_sensor referred to in these lines?
```
action:
    - delay: '00:01:00'
    - service: tesla_custom.polling_interval
      data:
        vin: "{{ state_attr('binary_sensor.' ~ trigger.entity_id.split('.')[1].split('_')[0] ~ '_online_sensor','vin') }}"
        scan_interval: 30
```

I have no sensor ending with `_online_sensor`. Has this perhaps been renamed?
I also have no entity called `tesla_custom.polling_interval`.

ddaddy:
Ah ok. I guess the `_sensor` ending has been removed.

So I have this which seems to make the HA UI happy:
```
action:
  - delay: "00:01:00"
  - service: tesla_custom.polling_interval
    data:
      vin: binary_sensor.jess_online
      scan_interval: 30
```

sgriessini:
You have to use the VIN attribute: _state_attr('binary_sensor.jess_online','vin')_

ddaddy:
Ah yes, that makes sense now.
Is there any way to view the current polling interval? I don't see a `tesla_custom.polling_interval` anywhere in the UI.

ddaddy:
> Ah ok. I guess the `_sensor` ending has been removed.
> 
> So I have this which seems to make the HA UI happy:
> 
> ```
> action:
>   - delay: "00:01:00"
>   - service: tesla_custom.polling_interval
>     data:
>       vin: binary_sensor.jess_online
>       scan_interval: 30
> ```

In case anyone else looks this up. This wasn't actually working as it needs to be written as"
```
  action:
    - service: tesla_custom.polling_interval
      data:
        vin: "{{ state_attr('binary_sensor.jess_online','vin') }}"
        scan_interval: 30
```

I would be good to have access to the polling interval through a sensor.

PR #948, opened 2024-04-24
feat: Expose polling interval as a sensor

ddaddy:
Exposes the polling interval as a sensor as requested in #946


ddaddy:
It seems the pre-commit is throwing Fatal Errors

alandtse (approved):

PR #953, opened 2024-04-26
perf: avoid writing state if poll did not happen

bdraco:
#948 made me realize that the car data was being polled far less frequently than I thought as the underlying library was caching, but every time the coordinator fired, it would still callback all the listeners and write the state of all the entities which meant we ended up writing state every 10 seconds even if nothing has changed.

Since the data the coordinator returns can be compared with `__eq__`, enabling `always_update=False` will check if the data has changed and only fire the listeners if it has. This avoids **thousands** of entity writes per minute which was the bulk of the state writes on my HA instance.  

This reduced the number of calls to `async_write_ha_state` by 62% on my production HA instance!

related PR https://github.com/home-assistant/core/pull/97268

### DIFF
--- a/custom_components/tesla_custom/__init__.py
+++ b/custom_components/tesla_custom/__init__.py
@@ -436,6 +436,7 @@ class TeslaDataUpdateCoordinator(DataUpdateCoordinator):
             _LOGGER,
             name=DOMAIN,
             update_interval=update_interval,
+            always_update=False,
         )
 
     async def _async_update_data(self):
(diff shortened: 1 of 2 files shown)


bdraco:
This is so significant that I'd almost want to call it a fix

bdraco:
Unfortunately it breaks the updates because the data returned via the coordinator is a boolean, and not the real data so it can't actually be compared.

However its nice to know that this integration is 62% of my state writes... so maybe that can be improved

PR #954, opened 2024-04-26
perf: avoid writing state when the controller did not actually update

bdraco:
https://github.com/alandtse/tesla/pull/953 didn't work because the data is actually a `bool`.  This is a second attempt

#948 made me realize that the car data was being polled far less frequently than I thought as the underlying library was caching, but every time the coordinator fired, it would still callback all the listeners and write the state of all the entities which meant we ended up writing state every 10 seconds even if nothing has changed.

Keep track of when the controller was last updated in each entity and if it has not changed, we skip the state write.

This reduced the number of calls to `async_write_ha_state` by 62% on my production HA instance!

### DIFF
--- a/custom_components/tesla_custom/base.py
+++ b/custom_components/tesla_custom/base.py
@@ -30,12 +30,6 @@ class TeslaBaseEntity(CoordinatorEntity[TeslaDataUpdateCoordinator]):
         self._attr_name = self.type.capitalize()
         self._attr_entity_registry_enabled_default = self._enabled_by_default
 
-    async def async_added_to_hass(self) -> None:
-        """Register state update callback."""
-        self.async_on_remove(
-            self.coordinator.async_add_listener(self.async_write_ha_state)
-        )
-
 
 class TeslaCarEntity(TeslaBaseEntity):
     """Representation of a Tesla car device."""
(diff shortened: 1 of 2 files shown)


bdraco:
This works as expected

bdraco:
need to push it to my other system for testing so I have it tested on two systems first

custom_components/tesla_custom/base.py
```diff
@@ -30,12 +30,6 @@ def __init__(
         self._attr_name = self.type.capitalize()
         self._attr_entity_registry_enabled_default = self._enabled_by_default
 
-    async def async_added_to_hass(self) -> None:
```
bdraco, line 32: This is already done in the BaseCoordinatorEntity 

https://github.com/home-assistant/core/blob/a25b2168a312713fafd73e769ab19f71a736ef6b/homeassistant/helpers/update_coordinator.py#L483

bdraco:
73 writes/min, down from 1000s

Screenshot 2024-04-26 at 12 40 25 PM

bdraco:
down to 78 on the other instance, nice!

Screenshot 2024-04-26 at 12 57 55 PM

alandtse:
Do we need to bump the minimum ha version too?

bdraco:
No need to bump the minimum version. It's more than new enough with this design

PR #965, opened 2024-05-11
feat: add bioweapon mode as fan mode

dandelionclock:
Added bioweapon mode as fan mode. Bioweapon mode can be turned on independently of other preset modes such as dog mode or camp mode, so it is better to include it as fan mode instead of one of the preset modes.
Requires [zabuldon/teslajsonpy/pull/468](https://github.com/zabuldon/teslajsonpy/pull/468)


alandtse:
Please handle the teslajsonpy bump and we can merge this in.

dandelionclock:
I updated the files with teslajsonpy 3.11.0. Not too sure about poetry.lock file but I updated the version and the hash manually

alandtse:
You need to run poetry update.

alandtse:
Thanks. Please take a look at the failing test. You can ignore the precommit error since that looks like something unrelated to climate.

dandelionclock:
I think I got it? Thank you for your patience.

PR #989, opened 2024-06-08
fix: update coordinator update time when receiving data from mqtt

### LINKED ISSUE #986: Location data not updating in HA via MQTT (or at least that's what it seems to be)

- [X] I have searched both the existing open issues & recently closed issues and did not find a duplicate of this issue.

### I have read about the Fleet API and understand I may need to use it

- [X] I understand issues relating to read only commands will be auto closed if not using the Fleet API.

v3.22.0

### Version of the Tesla car software

2024.14.9

### Model

Model Y

### Current Behavior

It seems that location data is not being polled via Teslamate MQTT.  I moved by Teslamate installation to a new machine and Teslamate itself is running fine. I can see my broker receiving data via MQTT. The Tesla Custom Integration used to be perfectly fine before this migration, but now it seems it won't pick any data via MQTT (specifically, location). I've even completely removed the integration and reinstalled it but nothing seems to work. The integration IS receiving data via MQTT - I can see that in debug logs - but it doesn't update the Home Assistant sensors based on that. My Garage Door was automated to open and close based on the location sensors and that takes almost 60 seconds now versus the almost instant behavior that I had come to expect. I'm completely confused here. 

### Expected Behavior

Sensors, especially location, should update in almost real-time because it's being polled via MQTT

### Debug logs

```true
Log File attached because it's too large to paste
```

### Anything else?

[home-assistant_tesla_custom_2024-06-03T00-46-18.324Z.log](https://github.com/user-attachments/files/15528428/home-assistant_tesla_custom_2024-06-03T00-46-18.324Z.log)

Log file attached because it's too large to paste otherwise

bdraco:
Note that I don't use mqtt so I don't have a way to test this.  I was only able to test that it didn't regress the incoming data from the coordinator polling.

fixes #986

### DIFF
--- a/custom_components/tesla_custom/__init__.py
+++ b/custom_components/tesla_custom/__init__.py
@@ -427,7 +427,7 @@ class TeslaDataUpdateCoordinator(DataUpdateCoordinator):
         self.update_vehicles = update_vehicles
         self._debounce_task = None
         self._last_update_time = None
-        self.last_controller_update_time: float | None = None
+        self.last_update_time: float | None = None
         self.assumed_state = True
 
         update_interval = timedelta(seconds=MIN_SCAN_INTERVAL)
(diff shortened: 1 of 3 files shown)


alandtse:
Thanks a bunch.

PR #990, opened 2024-06-08
perf: avoid creating a task for debounced listener update

bdraco:
Nothing was being awaited in _debounced so it could be a normal function decorated with `@callback` to run in the event loop and did not need to be a coro. This avoids creating a task when the async_call_later fires

Note as with #989, I don't use mqtt/teslamate so I have no real world test case for this

### DIFF
--- a/custom_components/tesla_custom/__init__.py
+++ b/custom_components/tesla_custom/__init__.py
@@ -6,6 +6,7 @@ from functools import partial
 from http import HTTPStatus
 import logging
 import ssl
+from typing import Any
 
 import async_timeout
 from homeassistant.config_entries import SOURCE_IMPORT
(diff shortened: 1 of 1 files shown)


custom_components/tesla_custom/__init__.py
```diff
@@ -504,17 +508,18 @@ def async_update_listeners_debounced(self, delay_since_last=0.1, max_delay=1.0):
...
-            self._debounce_task()
+        if self._cancel_debounce_timer:
+            self._cancel_debounce_timer()
             _LOGGER.debug("Previous debounce task cancelled")
 
-        # Schedule the call to _debounced, pass max_delay using partial
-        self._debounce_task = async_call_later(
-            self.hass, delay_since_last, partial(self._debounced, max_delay)
+        # Schedule the call to _async_debounced, pass max_delay using partial
+        self._cancel_debounce_timer = async_call_later(
```
bdraco, line 516: `async_call_later` does not return a task, it returns a callback that can be used to cancel the timer

Issue #977, opened 2024-05-29
Cooled Seats Setting


chrismcneil:
**Is your feature request related to a problem? Please describe.**
I only see options to turn on seat heating, it would be great if you could also turn on seat cooling.

**Describe the solution you'd like**
The dropdown menus for seat heating could also include seat cooling options.

**Describe alternatives you've considered**
I looked for seat cooling functionality outside of the seat heating options but could not find any.

**Additional context**
<img width="252" alt="image" src="https://github.com/alandtse/tesla/assets/3980643/3c02daa9-c11b-4ee4-9d01-a55ada39255c">

dandelionclock:
I'd like to attempt to add this, but I'll need help from OP or whoever because I don't have a Tesla with ventilated seats.

For starters, I'm trying to map out the API response values when seat coolers are on. Tesla API documentations are showing very helpful values of 0's across all seat climate states.

Can someone with ventilated seats set driver to low cool, front passenger to medium cool, rear left to high cool, rear center to auto, and rear right to high **heat** (for reference), then run the following service call in home assisant -> developer tools -> services -> yaml mode? Please replace <Put your VIN here, all CAPS> with your actual VIN, all in capital letters.

I'd like to see the response of climate_state, which for me are lines 73-112. I suspect the ventilated seats states are lumped with heated seats states under "seat_heater_left" "seat_heater_rear_center" "seat_heater_rear_left" "seat_heater_rear_right" and "seat_heater_right", just with different values (heated states are 0-4, so maybe cooled seats are 5-7?)

```
service: tesla_custom.api
data:
  command: VEHICLE_DATA
  parameters:
    path_vars:
      vehicle_id: <Put your VIN here, all CAPS>
```

chrismcneil:
Sorry the rear seats don't have cooling and the centre rear seat doesn't have anything in this 2024 m3, so I've done a couple combos below for you to reference.

Drivers Seat Cooling  level 3
Passenger Seat Cooling level 1
Left Rear Seat Heating level 3
Right Rear Seat Heating level 1
```
climate_state:
    allow_cabin_overheat_protection: true
    auto_seat_climate_left: false
    auto_seat_climate_right: false
    auto_steering_wheel_heat: true
    battery_heater: false
    battery_heater_no_power: null
    cabin_overheat_protection: "On"
    cabin_overheat_protection_actively_cooling: false
    climate_keeper_mode: "off"
    cop_activation_temperature: High
    defrost_mode: 0
[...]

PR #994, opened 2024-06-08
feat: add cooling seats

dandelionclock:
Bug reports are welcome because I cannot test extensively without a car with cooled seats.

[Issue 977](https://github.com/alandtse/tesla/issues/977)


alandtse:
Please handle merge conflicts.

alandtse:
So the way to handle merge conflicts is to pull in dev. You can typically do this by `git pull --rebase`. Manually making changes like you did doesn't really work well.

PR #995, opened 2024-06-09
fix: heated seats were not initialized if this integration is loaded when car is asleep

dandelionclock:
Integration was skip initialization of rear seats if it is loaded when car is asleep.


alandtse:
Please only target dev

dandelionclock:
Closing this for now. Need more work

Issue #725, opened 2023-10-06
Icons for climate presets


hanpq:
Now (as of HA 2023.10) with the tile card being able to show preset buttons it would be nice if icons could be added for the different presents. 

This is how it currently looks,
![image](https://github.com/alandtse/tesla/assets/33247576/46755c4d-9301-4ed4-bb4d-0eca26ea6ed5)

A suggestion of icon could be:

Normal = mdi:fan
Defrost = mdi:car-defrost-rear
Keep on = mdi:infinity
Dog mode = mdi:dog-side
Camp mode = mdi:tent

alandtse:
Thanks. Hopefully someone is willing to offer a PR to add the suggestions.

PR #1005, opened 2024-06-17
feat: add climate preset icons

dandelionclock:
Added icons for climate preset modes and fan modes

***Breaking Changes***
strings.json and icons.json must be used for adding icons to climate preset and fan modes. Translation keys in these files can only consist of lowercase letters, numbers, and underscore.

1. Preset modes has to be renamed to ["normal", "defrost", "keep", "dog", "camp"], and 
2. Fan modes has to be renamed to ["off", "bioweapon"].

Please update your scripts and automations after this update.

Fixes issue [725](https://github.com/alandtse/tesla/issues/725)

### DIFF
--- a/custom_components/tesla_custom/climate.py
+++ b/custom_components/tesla_custom/climate.py
@@ -50,6 +50,10 @@ class TeslaCarClimate(TeslaCarEntity, ClimateEntity):
     _attr_fan_modes = ["Off", "Bioweapon Mode"]
 
     @property
+    def translation_key(self):
+        return "car_climate"
+
+    @property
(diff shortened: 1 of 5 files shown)


alandtse:
Thanks. Breaking changes refer to changes that impact users. Dev changes aren't breaking.

dandelionclock:
> Thanks. Breaking changes refer to changes that impact users. Dev changes aren't breaking.

Although the UI didn't change, the states behind presets are now normal, defrost, keep, dog, and camp. Existing automations and scripts would probably still try to turn preset to "Dog Mode" which will now error out.

Issue #1029, opened 2024-08-10
Blocking call to load SSL certificates from disk


bdraco:
### Is there an existing issue for this?

- [X] I have searched both the existing open issues & recently closed issues and did not find a duplicate of this issue.

### I have read about the Fleet API and understand I may need to use it

- [X] I understand issues relating to read only commands will be auto closed if not using the Fleet API.

### Version of the Tesla component

v3.24.0

### Version of the Tesla car software

n/a

### Model

n/a

### Current Behavior

```
2024-08-10 10:25:42.645 WARNING (MainThread) [homeassistant.util.loop] Detected blocking call to load_default_certs with args (<ssl.SSLContext object at 0x7fbd8c7fbad0>, <Purpose.SERVER_AUTH: _ASN1Object(nid=129, shortname='serverAuth', longname='TLS Web Server Authentication', oid='1.3.6.1.5.5.7.3.1')>) inside the event loop by custom integration 'tesla_custom' at custom_components/tesla_custom/__init__.py, line 171: controller = TeslaAPI( (offender: /usr/local/lib/python3.12/ssl.py, line 713: context.load_default_certs(purpose)), please create a bug report at https://github.com/alandtse/tesla/issues
For developers, please see https://developers.home-assistant.io/docs/asyncio_blocking_operations/#load_default_certs
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/usr/src/homeassistant/homeassistant/__main__.py", line 223, in <module>
    sys.exit(main())
  File "/usr/src/homeassistant/homeassistant/__main__.py", line 209, in main
    exit_code = runner.run(runtime_conf)
  File "/usr/src/homeassistant/homeassistant/runner.py", line 190, in run
    return loop.run_until_complete(setup_and_run_hass(runtime_config))
  File "/usr/local/lib/python3.12/asyncio/base_events.py", line 674, in run_until_complete
    self.run_forever()
  File "/usr/local/lib/python3.12/asyncio/base_events.py", line 641, in run_forever
    self._run_once()
  File "/usr/local/lib/python3.12/asyncio/base_events.py", line 1992, in _run_once
    handle._run()
  File "/usr/local/lib/python3.12/asyncio/events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/src/homeassistant/homeassistant/setup.py", line 165, in async_setup_component
[...]

PR #1030, opened 2024-08-10
fix: fix api proxy cert blocking the event loop

bdraco:
load_verify_locations should run in the executor

related issue https://github.com/alandtse/tesla/issues/1029


custom_components/tesla_custom/__init__.py
```diff
@@ -181,7 +184,6 @@ async def async_setup_entry(hass, config_entry):
             polling_policy=config_entry.options.get(
                 CONF_POLLING_POLICY, DEFAULT_POLLING_POLICY
             ),
-            api_proxy_cert=config.get(CONF_API_PROXY_CERT),
```
bdraco, line 184: No point in passing this if we already loaded it and we pass the AsyncClient

PR #1067, opened 2024-11-22
fix: Bump teslajsonpy to 3.12.2

ehendrix23:
Bump teslajsonpy to 3.12.2 which includes fix for API service


alandtse:
Please review other bumps. Other files will also also need to be changed.

ehendrix23:
Updated the additional files. Not sure why the test would fail though as nothing there has been changed.

alandtse:
Someone probably broke the test with a commit.

alandtse:
Thanks for figuring that out. Is the other PR #1069 a prereq of this? Normally I'd want to keep the test fix as a separate PR so the bump PR is very clean.

ehendrix23:
> Thanks for figuring that out. Is the other PR #1069 a prereq of this? Normally I'd want to keep the test fix as a separate PR so the bump PR is very clean.

Nope, no dependencies. PR #1069 was purely to update for HA 2024.11 and had nothing to do with teslajsonpy. :-)

FYI, seems I screwed some things up so closing this PR for the bump and creating a new one instead that is clean. :-)

custom_components/tesla_custom/services.py
```diff
@@ -5,11 +5,11 @@
 
 import logging
 
+import voluptuous as vol
```
alandtse, line 8: Did black do this?

ehendrix23: I would say yes, I didn't add it. :-)

Issue #1092, opened 2025-02-04
Migrating from Polling to Fleet Telemetry for Cost Efficiency


flaviorighi:
According to the [Tesla developer site](https://developer.tesla.com/docs/fleet-api/fleet-telemetry), continuously polling the `vehicle_data` endpoint is costly and inefficient. 
Instead, it's recommended to transition to using the Telemetry service.

By migrating to a setup that uses Fleet-Api for sending commands and Telemetry for gathering data, we can achieve significant improvements, particularly for users managing multiple Tesla vehicles through the same Home Assistant instance.

After last year fleet migration, each HA instance is already need to be exposed to the internet, transitioning to the Telemetry service should be a straightforward process. This change would help reduce costs and improve efficiency for all users relying on this integration.

DeadDjembe:
This would be great, Tesla started charging for polling in Feb.  Each user receives $10 of credit per month, but I chewed threw the credit in a just few days.  Currently, this integration polls driving cars every 60 seconds and this is not configurable. This will eat up the credit quickly.  

Ultimately it would be ideal to integrate Telemetry into this integration, but for the meantime it would be nice to have the option to turn off automatic polling all together (including while driving) but allow the option to manually poll.  I have the Telemetry service integrated into MQTT that works very well and is updated nearly instantaneously, so I can get the data into HA.  But, this integration is required to send the command.  Writing a script to manually poll when something interesting happens in the Telemetry MQTT is an easy work around to keep under the free limit and limit the number of polls. 

See here for more info about the new billing, there is a section about Cost Optimization that discusses integrating the Telemetry service in place of continually polling:  https://developer.tesla.com/docs/fleet-api/billing-and-limits

flaviorighi:
Yeah, probably I missed the main reason because I wrote that, but @DeadDjembe explained very well the point.

After opened this issue I've took a look deeper at the code and at the Tesla's API calls graph and I figured out that the problem is all about the 60 sec polling while driving.
But they doesn't have control on that here, cause the hardcoded polling value is on the teslajsonpy side.

So since it would be a faster change to have in our HA sistems, I've ended opening [another issue](https://github.com/zabuldon/teslajsonpy/issues/481#issue-2828971544) on teslajsonpy library.

BogdanDIA:
It will be good to have the telemetry-api integrated. However, for now I created my own add-on in HA that send commands through bluetooth. It works great for charging from solar surplus.

rubin110:
[...]

PR #1095, opened 2025-03-09
chore: upgraded python to 13.3 support for 13.2.8 ends in HA 2025.2

flaviorighi:
Updated the project python version to the minimum supported by HA 2025.2


flaviorighi:
@alandtse this PR is just necessary to open another PR related to #1092, I've already wrote the code to update the new version of teslajsonpy in this project.

I'm waiting this PR to be accepted before to open the real one.

alandtse:
Your diffs are huge. Figure out why you're settings are touching so much of the file please.

alandtse:
Like I'm ok with the poetry.lock diff being huge. But I have no idea why the pyproject.toml change is basically the whole file.

flaviorighi:
I didn't check it yet, I think it could be just the line separator, since I used WSL on windows it can be happened.

rubin110:
Would be super cool if this change went through. Switching over to telemetry seems like the future.

alandtse:
Sorry lost track of this. Manually made these changes.

PR #1098, opened 2025-03-21
feat: add support for 'Nothing' as a Powerwall export selection

ChrisJ60:
Add support for specifying that the Powerwall system should not export any energy to the grid.


alandtse:
The test looks broken. Please check. I don't know who setup this mock test so not sure what it's trying to do. I don't have a powerwall so this means nothing to me.

tests/test_select.py
```diff
@@ -343,6 +343,17 @@ async def test_energy_exports(hass: HomeAssistant) -> None:
     with patch(
         "teslajsonpy.energy.SolarPowerwallSite.set_export_rule"
     ) as mock_set_export_rule:
+        # Test selecting "Nothingr"
```
alandtse, line 346: Also typo here.

ChrisJ60:
I fixed the typo. This test file seems to be testing all of the different 'Select' controls, most of which seem to refer to cars. If the test is actually broken then it was broken before my addition. I simply inserted these extra lines:

        # Test selecting "Nothing"
        await hass.services.async_call(
            SELECT_DOMAIN,
            SERVICE_SELECT_OPTION,
            {
                ATTR_ENTITY_ID: "select.battery_home_energy_exports",
                "option": "Nothing",
            },
            blocking=True,
        )
        mock_set_export_rule.assert_awaited_once_with("never")

which exactly follow the existing pattern but adds a test for the new export option (Nothing/never). This specific test is testing the (was 2, now 3) possible settings for energy export to the grid, all of which work fine in production.

As I am unable to run the tests, and there seems little prospect that I will ever be able to, I'm not able to check this any further.

alandtse:
Unfortunately, the failing test is the one you modified. As the person most likely to fix the failing CI test is the person who added the test that broke it. People with totally unrelated code won't have any urge to help fix it. I'll leave this unmerged until you or someone else fixes it.  

I'm not sure of the purpose for `assert_awaited_once_with` and whether there's a non_once variant that would resolve it here or if it makes sense to break out the test for each option instead of for all options at once.

alandtse:
I would suggest just duplicating the select test and changing for your added parameters since it looks like it's something that changed with your added element.

ChrisJ60:
Not sure I follow exactly what you mean... Do you mean add that one select option as a separate test in the current file? or create an entirely new test file with just that one option test in it? or...?

Can you confirm that (a) when I restored the original test file the tests ran without issue? And (b) when the test was failing what exactly were the symptoms, errors etc.? (since I'm not able to run the tests myself).

alandtse:
Tests are separate functions in the same file. I was suggesting creating a new function to test your new option. And yes, the ci is fine without your test so it broke due to what you added.

ChrisJ60:
Okay. As I asked though, are there any logs of the (broken) test output that might give me a clue as to why it broke, since there is nothing at all obvious as to why that simple addition should break it. Without something to go on I'm kind of just floundering in the dark.

alandtse:
I don't believe the prior log runs are reviewable after a new run completes. The error was the assertion happened more than once. The test only allows for one assertion.

skipishere:
It's still viewable here https://github.com/alandtse/tesla/actions/runs/14030023396/job/39303981195

ChrisJ60:
Aha, that helps a lot. Looking carefully at the original test (which I think has an issue anyway) I think I see the problem. I have tweaked that and also added my additional test back in as I think it should be. Let's see what the next test run shows.

alandtse:
Looks like that was it. Thanks for handling.

PR #1128, opened 2025-07-20
fix: Add try-catch logic around powerwall reserve update

### LINKED ISSUE #1125: call to set backup reserve level fails; errors in HA log

- [x] I have searched both the existing open issues & recently closed issues and did not find a duplicate of this issue.

3.25.0

### Version of the Powerwall Gateway software

25.18.4 b6b41ca8

### Model

2

### Current Behavior

(Very) occasionally a call to set the powerful backup reserve level fails and generates errors in the HA log. This aborts the automation that is running which then leads to other consequences. This same code is executed every day at the same time and in 99.9% of cases it works just fine but occasionally this failure occurs.

### Expected Behavior

 I don't know what causes the failure but I think it needs to be handled better to allow the automation that calls this function to continue rather than be aborted.

### Debug logs

```true
Since the issue is very infrequent I don;t have debug logs, but here are the relevant messages from the HA log when the error occurred last night. The value being passed (i.e. the value to set the reserve to) is, in this case, 100.

2025-07-19 04:00:00.701 ERROR (MainThread) [homeassistant.components.automation.pw_force_charge_set_backup_reserve_to_100_at_23_30] PW OFF-PEAK CHARGING: Set Backup Reserve to maximum off-peak charge level during off-peak period: If at step 2: If at step 1: If at step 1: Error executing script. Unexpected error for call_service at pos 1: 'code'
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 524, in _async_step
    await getattr(self, handler)()
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 1012, in _async_step_call_service
    response_data = await self._async_run_long_action(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<9 lines>...
    )
    ^
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 624, in _async_run_long_action
    return await long_task
           ^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2799, in async_call
    response_data = await coro
                    ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2842, in _execute_service
    return await target(service_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 1011, in entity_service_call
    single_response = await _handle_entity_call(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
        hass, entity, func, data, call.context
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 1083, in _handle_entity_call
    result = await task
             ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/number/__init__.py", line 122, in async_set_value
    await entity.async_set_native_value(native_value)
  File "/config/custom_components/tesla_custom/number.py", line 125, in async_set_native_value
    await self._energysite.set_reserve_percent(value)
  File "/usr/local/lib/python3.13/site-packages/teslajsonpy/energy.py", line 219, in set_reserve_percent
    if data and data["response"]["code"] == 201:
                ~~~~~~~~~~~~~~~~^^^^^^^^
KeyError: 'code'
```

### Anything else?

_No response_

skipishere:
Attempts to hack-fix #1125 until teslajsonpy is updated to address API changes.

Currently the call to Tesla to update the Powerwall reserve succeeds but it no longer contains a `code` element and as the teslajsonpy expects this it exceptions which bubbles up and causes automations using this to error out (when in reality it has succeeded). There is an [issue raised to address](https://github.com/zabuldon/teslajsonpy/issues/486) this but in the mean time this hack fix should unblock this repo & can be reverted (or improved) once the dependency is sorted out.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when setting backup reserve values by handling errors silently, ensuring Home Assistant state updates continue even if an error occurs.

### DIFF
--- a/custom_components/tesla_custom/number.py
+++ b/custom_components/tesla_custom/number.py
@@ -122,8 +122,13 @@ class TeslaEnergyBackupReserve(TeslaEnergyEntity, NumberEntity):
 
     async def async_set_native_value(self, value: int) -> None:
         """Update backup reserve percentage."""
-        await self._energysite.set_reserve_percent(value)
-        self.async_write_ha_state()
+        try:
+            await self._energysite.set_reserve_percent(value)
+        except:
+            # Hack to stop exception blocking automation until teslajsonpy is updated
+            pass
+        finally:
+            self.async_write_ha_state()
 
     @property
     def native_value(self) -> int:


alandtse:
I'd prefer that we get a fix into teslajsonpy and bump that instead of a blanket exception swallow. Is the real fix too hard to put together?

skipishere:
@ChrisJ60 has made a PR https://github.com/zabuldon/teslajsonpy/pull/487 to address this, will close this as that fix is the preferred route

PR #1149, opened 2025-10-11
fix: powerwall battery sensor loading (#1148)

alandtse:


me-bhavyaaa:
hey, I was having some issues with the Alexa media player. How can I get in contact with you?

PR #1187, opened 2026-05-16
fix: ignore stale TeslaMate VIN mappings

oformaniuk:
## Summary
- Fix TeslaMate MQTT car-id lookup so stale VIN mappings do not block the active loaded VIN for the same TeslaMate id
- Continue checking other VIN mappings when a mapped VIN is not loaded
- Add regression coverage for stale VIN mappings appearing before the active VIN

This fixes the integration's TeslaMate MQTT consumer when multiple stored VIN mappings point at the same TeslaMate id. TeslaMate itself is publishing retained MQTT messages correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed car lookup to properly skip stale vehicle mappings and identify the correct vehicle.

* **Tests**
  * Added test coverage for handling stale vehicle mappings in car identification.

[Review Change Stack](https://app.coderabbit.ai/change-stack/alandtse/tesla/pull/1187?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/custom_components/tesla_custom/teslamate.py
+++ b/custom_components/tesla_custom/teslamate.py
@@ -212,20 +212,21 @@ class TeslaMate:
 
         await self.async_load()
 
-        found_vin = None
-
         car_map = self._data.get("car_map", {})
         for vin, tm_id in car_map.items():
-            if tm_id == teslamate_id:
-                found_vin = vin
-                break
+            if tm_id != teslamate_id:
+                continue
 
-        if found_vin is None:
-            return None
+            if car := self.cars.get(vin):
+                return car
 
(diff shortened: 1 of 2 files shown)


alandtse:
I don't use teslamate, but would this impact users with multiple vehicles negatively?

alandtse:
@oformaniuk any thoughts on my question?

oformaniuk:
Hey, sorry I've missed your question earlier.
It should not have negative impact.
Reason for the fix was to unblock my use case as I got new car and was not able to make it work due to VIN mismatch.

PR #1216, opened 2026-06-20
feat: add powerwall mode select for charging

### LINKED ISSUE #1169: Charge from Grid / Discharge / Hold (use Grid)

- [x] I have searched both the existing open issues & recently closed issues and did not find a duplicate of this issue.

3.25.4

### Version of the Powerwall Gateway software

25.42.1

### Model

2

### Current Behavior

This seems to be more a Tesla Powerall Gateway issue, but I ask it here, because, people use this integration to control the powerwall.

As a result of the ha-wattwise hems, I need to have three states of the powerwall: charging from grid, discharging, keep charge (supply home from grid). 

Right now, exactly for the keep charge state I have not found a way to enforce it: Charge from Grid switch could be off and on, it does not care.

If I set a tarif plan (which is only possible via Tesla App), then the powerwall (or Tesla?) decides when to charge and when not. If I set to self_consume and raise the reserve higher than the current level, it starts to charge from grid...

Any hints what else I could try or what could work? I guess it would work to set the tarif dynamically via the API... but this is not possible or is it?

### Expected Behavior

Three states: Charge from Grid / Discharge / Hold (use Grid)

### Debug logs

```true
no logs right now.
```

### Anything else?

_No response_

compiledbyutkarsh:
Fixes #1169

## Summary
Users needed a single control to manage Powerwall charging behavior combining `operation_mode` and `grid_charging` settings together.

## Changes
- Added new `TeslaEnergyPowerwallMode` select entity
- Three states supported:
  - **Charge from Grid** → `operation_mode: self_consumption` + `grid_charging: True`
  - **Discharge** → `operation_mode: self_consumption` + `grid_charging: False`
  - **Hold (use Grid)** → `operation_mode: autonomous` + `grid_charging: True`
- Added tests for all three states and registry entry

## Summary by CodeRabbit

* **New Features**
  * Added a “Powerwall mode” select for battery energy sites (Charge from Grid, Discharge, Hold (use Grid)), mapping selections to the underlying grid-charging and operation settings.

* **Bug Fixes**
  * Improved power and distance sensors to treat non-finite numeric, invalid, or unconvertible values as unknown.
  * Updated relevant select entities to return no selection (`None`) when the current value doesn’t match any known option.

* **Tests**
  * Extended select tests to verify the new Powerwall mode entity and its option-to-settings behavior.
  * Updated/added sensor tests for `None` and edge-case validation.

### DIFF
--- a/custom_components/tesla_custom/select.py
+++ b/custom_components/tesla_custom/select.py
@@ -3,6 +3,7 @@
 import logging
 
 from homeassistant.components.select import SelectEntity
+from homeassistant.exceptions import HomeAssistantError
 from homeassistant.core import HomeAssistant
 from homeassistant.helpers.entity import EntityCategory
 from teslajsonpy.car import TeslaCar
@@ -415,7 +416,7 @@ class TeslaEnergyPowerwallMode(TeslaEnergyEntity, SelectEntity):
             return POWERWALL_MODE[1]
         if op_mode == "autonomous" and grid_charging:
             return POWERWALL_MODE[2]
-        return ""
+        return None
 
 
 class TeslaEnergyGridCharging(TeslaEnergyEntity, SelectEntity):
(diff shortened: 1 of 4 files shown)


alandtse:
Please resolve all AI comments.

